PR #24, opened 2018-08-16
Update skaffold.yaml to add gcb build

ocervell:
`skaffold run` will fail while building the `cartservice` with a 'No Space Left on Device Error'
Increasing the disk allocated for the build (default 200GB) to get rid of this error.

### DIFF
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -17,8 +17,6 @@ kind: Config
 build:
   tagPolicy:
     gitCommit: {}
-  googleCloudBuild:
-    diskSizeGb: 300
   artifacts:
   - imageName: gcr.io/microservices-demo-app/emailservice
     workspace: src/emailservice
@@ -44,3 +42,9 @@ deploy:
   kubectl:
     manifests:
     - ./kubernetes-manifests/**.yaml
+profiles:
+  - name: gcb
+    build:
+      googleCloudBuild:
+        diskSizeGb: 300
+        machineType: "N1_HIGHCPU_32"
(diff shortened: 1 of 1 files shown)


ahmetb:
Adding this change forces the skaffold.yaml to force running on GCB which is not something we want.

Instead, can you please do:

```yaml
# You can also use Google Cloud Builder to build
# the images on the cloud with "skaffold run -p gcb".
profiles:
  gcb:
    build:
      googleCloudBuild: [...]
```

so that others not using GCB aren't effected?

ocervell:
Done, better this way indeed. 

Should we mention the build VM disk space needed somewhere in the docs ?

ahmetb:
I think this would be a problem if you're building the cartservice many times, so you have unused layers of it laying around to consume a lot of space. I reckon you're not actively refactoring the cartservice –so not sure why this is happening to you.

I don't understand why ~20 GB (or whatever the GCB default is) isn't enough to build all images once. I've developed most of this demo from scratch on 64GB "docker for mac" with many stale layers laying around for each image, and never ran out of disk.

Is your concern primarily "laptop disk space" or "GCB build machine disk space"?

ahmetb:
I have annotated the GCB section of the yaml, now merging.

We won't be using this for now, but I'm ok with keeping it to showcase GCB.

ahmetb:
Overriding CLA check caused by creating a commit (by me) from the GitHub web UI.

dzuluaga:
This issue helped me understand why I was getting "No Space Left on Device Error" on this tutorial. Thanks for putting this together!

PR #28, opened 2018-08-21
grpc: add health checks to python services

ahmetb:
also converted line endings for recommendationservice/requirements.txt from
dos to unix.


ahmetb:
fyi @DanSanche @michaelawyu

PR #30, opened 2018-08-24
Instrument Go service with Stats and monitored resources.

rghetia:
added default views for stats  for http and grpc in Go services only.
added monitored resources for go.


rghetia:
Screenshots of stats after instrumentation.
https://screenshot.googleplex.com/ssh5rYoHgs1
https://screenshot.googleplex.com/DzbZ3tNOE3N

PR #33, opened 2018-08-27
k8s/adservice: bump memory limits for adservice

ahmetb:
adservice has beeing OOMkill'ed over the weekend. it looks like the memory
goes up to ~130MiB and then gets killed.

image


ahmetb:
cc: @rghetia

Issue #40, opened 2018-08-31
Consider using minikube instead of Docker for Desktop for the local setup


idvoretskyi:
I'd suggest using [minikibe](https://github.com/kubernetes/minikube) instead of Docker for Desktop for the local setup (https://github.com/GoogleCloudPlatform/microservices-demo/#option-1-running-locally-with-docker-for-desktop).

Minikube is the vendor-neutral tool, driven by the Kubernetes community, that also offers cross-platform compatibility (eg. you can't use Kubernetes with Docker for Desktop for Linux).

PR #41, opened 2018-08-31
Minikube instructions added

idvoretskyi:
Fixes https://github.com/GoogleCloudPlatform/microservices-demo/issues/40

Signed-off-by: Ihor Dvoretskyi <ihor@linux.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Find **Protocol Buffers Descriptions** at the [`./pb` directory](./pb).
 ## Features
 
 - **[Kubernetes](https://kubernetes.io)/[GKE](https://cloud.google.com/kubernetes-engine/):**
-  The app is designed to run on Kubernetes (both locally on "Docker for
+  The app is designed to run on Kubernetes (both locally on Minikube and "Docker for
   Desktop", as well as on the cloud with GKE).
 - **[gRPC](https://grpc.io):** Microservices use a high volume of gRPC calls to
   communicate to each other.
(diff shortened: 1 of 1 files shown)


ahmetb:
I am ok with removing Docker-for-Desktop altogether! Let’s just talk about Minikube; adding more options just confuse people.

ahmetb:
Thanks Ihor!

I have two questions:

1. Have you built this yourself on minikube? I ask this because sometimes the disk space is too small for building all images.

2. Is the app accessible on localhost:80 with minikube? (My assumption is: No.) Therefore this part of the docs might be invalid now:

    > The    application frontend should be available at http://localhost:80 on your     machine.

README.md
```diff
@@ -91,7 +88,7 @@ Find **Protocol Buffers Descriptions** at the [`./pb` directory](./pb).
 
 > 💡  Recommended for demos and making it available publicly.
 
-1. Install tools specified in the previous section (Docker, kubectl, skaffold)
+1. Install tools specified in the previous section (kubectl, skaffold)
```
ahmetb: This part of the diff is not accurate: Skaffold will still need a working `docker build` locally even if it is deploying to a remote cluster.

I'm tempted to think: for now Docker for Desktop is doing a decent job here and we should probably keep it.

ahmetb:
I just did

    minikube start --vm-driver=xhyve --cpus=3 --memory 4096
    skaffold dev

and this brought the apiserver down or something like that, because `kubectl`  to minikube now fails with:

    The connection to the server 192.168.64.29:8443 was refused - did you specify the right host or port?

and it used to work. :(

idvoretskyi:
@ahmetb wow, thanks for testing. Will take a look l and update the instructions respectively.

README.md
idvoretskyi: @ahmetb It's fine with keeping DfD, the primary reason why I've added minikube - it's available for Linux.

idvoretskyi:
@ahmetb it really doesn't work on minikube, and that's weird.

Let's hold the PR now, I'll perform a deeper debugging and update the PR.

ahmetb:
I bumped the limits but no luck. If you have time, feel free to follow up on Minikube repo. I'm still open to this as long as (1) it works (2) localhost:80 or something like that is accessible.

idvoretskyi:
@ahmetb I'll close this PR now. We have an issue https://github.com/GoogleCloudPlatform/microservices-demo/issues/40 opened, where we may track progress.

I'll submit another PR after I'll figure out a workable way to run this on Linux locally.

dreamspider42:
Hey guys,
I found this link helpful: https://stackoverflow.com/questions/40767164/expose-port-in-minikube
I was able to finally able to use the example successfully after getting the minikube ip and then hitting the appropriate port such as http://192.168.39.59:30979.  I'm still brand new to Kubernetes and this solution, but do think it would be worthwhile offering folks some clear instructions on how to do this in linux if they're looking for it.  After some trial and error I got it working locally using the following steps:
1. Install MiniKube according to the install instructions [here](https://github.com/kubernetes/minikube)
1. Install the Linux KVM2 driver according to the instructions [here](https://github.com/kubernetes/minikube/blob/master/docs/drivers.md#kvm2-driver)
1. Set the default vm driver to KVM2 with the following command: `sudo minikube config set vm-driver kvm2`
1. Install KubeCtl via the following command: `sudo snap install kubectl --classic`
1. CHOWN the minikube directory with the following command: `sudo chown -R $USER $HOME/.minikube` and `sudo chown -R $USER $HOME/.kube`

Interested if it works for anyone else as well.

dreamspider42:
Also had an issue with skaffold trying to push to docker hub so added this and it seemed to resolve: `skaffold config set -k minikube local-cluster true`

idvoretskyi:
@dreamspider42 It works since minikube v1.0, I'll update the documentation when I'll have time.

idvoretskyi:
@dreamspider42 also, this is not necessary:

> CHOWN the minikube directory with the following command: sudo chown -R $USER $HOME/.minikube and sudo chown -R $USER $HOME/.kube

Instructions to run Minikube from this blog post are sufficient - https://kubernetes.io/blog/2019/03/28/running-kubernetes-locally-on-linux-with-minikube-now-with-kubernetes-1.14-support/.

dreamspider42:
Yeah, its possible I screwed something up during the minikube install as I couldn't run minikube without sudo and then it wouldn't show me the dashboard.  Thanks for the link though.  I think the bigger thing is I wasnt sure where to hit the front-end URL via minikube (it wasn't available via localhost:80) without the commands in the other link.

PR #44, opened 2018-09-20
Introducing super basic health check for cart service

simonz130:
- Generated C# proto implementation for grpc health check
  - Moved all C# protos to a dedicated folder
  - Implemented basic health checking to ping CartStore (which is Redis in default implementation)
  - Base plumbing for health checks


src/cartservice/cartstore/RedisCartStore.cs
```diff
@@ -197,5 +197,18 @@ public async Task<Hipstershop.Cart> GetCartAsync(string userId)
                 throw new RpcException(new Status(StatusCode.FailedPrecondition, $"Can't access cart storage. {ex}"));
             }
         }
+
+        public bool Ping()
+        {
+            try
+            {
+                EnsureRedisConnected();
```
ahmetb: I think current impl of EnsureRedisConnected always returns true after establishing connection once. We probably need to actually make a request to redis here, like https://redis.io/commands/ping.

simonz130: It returns **true** if connection is established, but if connection fails, the callback *ConnectionFailed* will be called, which will set *isRedisConnectionOpened to **false**, so it's possible that subsequent execution of the method will get to 
```
                if (redis == null || !redis.IsConnected)
                {
                    Console.WriteLine("Wasn't able to connect to redis");
                    
                    // We weren't able to connect to redis despite 5 retries with exponential backoff
                    throw new ApplicationException("Wasn't able to connect to redis");
                }
```
which which throw exception that will be reflected in a health probe returning **false**.

It's a bit complex logic - maybe I'll simplify it. Also, I used 
```
                Console.WriteLine("Performing small test");
                cache.StringSet("cart", "OK" );
                object res = cache.StringGet("cart");
                Console.WriteLine($"Small test result: {res}");
```
for *ping* analog, so I'll update it to *Ping* API call on redis.

Issue #47, opened 2018-09-24
JSON Structured logs for Stackdriver Logging


ymotongpoo:
Currently services in this demo, at least those written in Go, are emitting logs to stderr as of now, which makes the appearance of logs confusing because some info level logs are treated as error level ones, due to Stackdriver Logging's default setting. Changing the log destination to stdout solves this.

Also, all log messages are in text payload. Though this is the demonstration of GKE, it would be great if it can demonstrate the features of Stackdriver Logging as well. For that, using JSON (structured log format) can show how Stackdriver Logging can handle the log more efficiently such as handy filters.

## Progress tracker

(added by ahmetb, cc: @ymotongpoo  )

- [x] adservice (#59)
- [ ] cartservice
- [x] checkoutservice (#48)
- [x] currencyservice (#66)
- [x] emailservice (#66)
- [x] frontend
- loadgenerator (not eligible, produces human-readable output)
- [x] paymentservice (#66)
- [x] productcatalogservice (#48)
- [x] recommendationservice (#66)
- [x] shippingservice (#48)

askmeegs:
cartservice is the only service remaining to complete this -- closing in favor of https://github.com/GoogleCloudPlatform/microservices-demo/issues/154

PR #48, opened 2018-09-24
Change log format to JSON and log destination to stdout

ymotongpoo:
This is the PR for #47. This PR effects all the services written in Go.

The summary of this PR is:
1. Change log format from text payload to JSON payload.
2. Change log destination from stderr to stdout.
3. Change target container repository to mine so that I can test changes from 1 and 2.
4. Change scenario of locust test.

For 3 and 4, I'll revert them to the original once the objective is approved.


ahmetb:
> Change log destination from stderr to stdout.

I'm not sure if stderr tells anything about the log to Stackdriver. In my https://github.com/ahmetb/coffeelog project I've seen that output stream doesn't matter as long as you get the Stackdriver json fields right. 

As for (3) and (4), not sure why you felt like you did "git add" these files. 😄 You can test those without adding them to your commit.

src/shippingservice/quote.go
```diff
@@ -50,6 +50,6 @@ func quoteByCountFloat(count int) float64 {
 		return 0
 	}
 	count64 := float64(count)
-	var p float64 = 1 + (count64 * 0.2)
+	var p = 1 + (count64 * 0.2)
```
ahmetb: I'm not sure what this line changes. isn't it already float64?

ymotongpoo: `var p` doesn't require `float64` type definition as it's inferred from `count64`. Moreover now I realized it doesn't require `var` keyword as well.

`p := 1 + (count64 * 0.2)` just works.

PR #55, opened 2018-09-27
adservice: use the Stackdriver Logging library and OpenCensus log correlation

sebright:
I was experimenting with using OpenCensus log correlation in the ad service, so I made a pull request in case that is something that would be useful to include in the demo.  This PR depends on #54.  The other commits are:

#### adservice: export logs with the Stackdriver Logging library

This commit exports logs from the adservice directly to Stackdriver using the
Stackdriver Logging library.  Using the Stackdriver Logging library instead of
unstructured logging gives more control over setting fields such as severity.
This change can also be used to enable OpenCensus log correlation
(https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/log_correlation/stackdriver).

#### adservice: enable OpenCensus log correlation

OpenCensus log correlation
(https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/log_correlation/stackdriver)
inserts the trace ID, span ID, and sampling decision into every log entry.  The
tracing data can be used to find all log entries associated with a single trace
or span.

This commit also upgrades OpenCensus to 0.16.1.

_____________________________________________________________

Here is a screenshot of a trace and an associated log entry:

log-correlation

/cc @rghetia @g-easy

### DIFF
--- a/src/adservice/build.gradle
+++ b/src/adservice/build.gradle
@@ -54,7 +54,8 @@ dependencies {
                 "io.grpc:grpc-services:${grpcVersion}",
                 "io.prometheus:simpleclient_httpserver:${prometheusVersion}"
 
-        runtime "io.opencensus:opencensus-impl:${opencensusVersion}",
+        runtime "com.google.cloud:google-cloud-logging:1.45.0",
+                "io.opencensus:opencensus-impl:${opencensusVersion}",
                 "io.netty:netty-tcnative-boringssl-static:2.0.8.Final"
     }
 }
@@ -106,6 +107,7 @@ task adService(type: CreateStartScripts) {
     applicationName = 'AdService'
     outputDir = new File(project.buildDir, 'tmp')
     classpath = jar.outputs.files + project.configurations.runtime
+    defaultJvmOpts = ["-Djava.util.logging.config.file=./src/main/resources/logging.properties"]
 }
 
 task adServiceClient(type: CreateStartScripts) {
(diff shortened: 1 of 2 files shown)


ahmetb:
Thanks for the patch. Some questions since I'm not familiar with the particular libraries:

1. Does this still log messages to stderr or stdout on GCP?
2. Does this still work (program doesn't crash/fail) if I don't provide any GCP credentials (running on a laptop or AWS)?

sebright:
I didn't realize that the demo also needed to run without access to Stackdriver APIs.  I think that this PR currently doesn't print the logs to stdout, and I'm not sure how to use google-cloud-logging when Stackdriver is available and use a ConsoleHandler when it isn't available.

Another option is to use opencensus-contrib-log-correlation-log4j2 for log correlation, which doesn't require google-cloud-logging.  I could change the log statements to use Log4j, continue printing the log to stdout, and specify a JSON format (with tracing fields) that could be read by the Stackdriver Logging agent.  Do you think that configuration would work for this project?

ahmetb:
Sorry for not providing the context above earlier. We have not documented them before. I've outlined some development principles for this repo here: #56.

sebright:
Thanks for explaining the requirements and adding the documentation!  Do you think that my idea for using Log4j above would work?  The only change in behavior would be outputting structured logs, which would contain tracing fields that are understood by the Stackdriver Logging agent, instead of unstructured logs.

ahmetb:
> outputting structured logs, which would contain tracing fields that are understood by the Stackdriver Logging agent, instead of unstructured logs.

Yes, if all the change is just printing structured logs recognized by SD Logging, that works. We recently merged a PR that converted all Go services to structured JSON logs.

I don't know anything about log4j so I'll defer to you or `@`rghetia.

sebright:
@ahmetb Thanks!  I made a new PR (#59), so I'll close this one.

PR #60, opened 2018-09-28
pb: add "categories" field to Product

sebright:
This field can be used as the context keys to look up relevant ads in the ad
service.

/cc @rghetia 

I also ran the genproto.sh scripts for the Java and Go services and included those changes in the second commit.  I encountered an issue when I ran genproto.sh for the recommendation service, and I'm still looking into it.

PR #61, opened 2018-09-29
adservice: find relevant ads by category

sebright:
The ad service now returns ads matching the categories of the product that is
currently displayed.  Changes in this commit:

- List all products' categories in products.json.

- Pass the current product's categories from the frontend to the ad service when
  looking up ads.

- Store a statically initialized multimap from product category to ad in the ad
  service.

- Return all ads matching the given categories when handling an ads request.
  The ad service continues to return random ads when no categories are given or
  no ads match the categories.

________________________________________________________________________

~~This PR depends on #60 being merged into master and src/frontend/Gopkg.lock being updated to reference the new commit.~~

/cc @rghetia 

Do you think it makes sense to initialize the ads map statically and make it immutable as in this PR, or will it need to be updated at runtime?

### DIFF
--- a/src/frontend/Gopkg.lock
+++ b/src/frontend/Gopkg.lock
@@ -27,17 +27,6 @@
   version = "v0.5.0"
 
 [[projects]]
-  branch = "master"
-  digest = "1:3ef905a7059a17712b7b27315692992f84f356e828d38f6ff0624e04103ec675"
-  name = "github.com/GoogleCloudPlatform/microservices-demo"
-  packages = [
-    "src/frontend/genproto",
-    "src/frontend/money",
-  ]
-  pruneopts = "UT"
-  revision = "86c8c06cc1e0aa616796a19493f042466d93098d"
-
-[[projects]]
   digest = "1:72856926f8208767b837bf51e3373f49139f61889b67dc7fd3c2a0fd711e3f7a"
   name = "github.com/golang/protobuf"
   packages = [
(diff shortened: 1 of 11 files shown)


rghetia:
I don't expect that we would update Ad mapping at runtime. So immutable map is fine.

src/frontend/Gopkg.lock
```diff
@@ -35,7 +35,7 @@
     "src/frontend/money",
   ]
   pruneopts = "UT"
-  revision = "6d969441585ade8c91c235115c7cdb12ac61354f"
+  revision = "86c8c06cc1e0aa616796a19493f042466d93098d"
```
ahmetb: actually this is bad. I'm not sure why the repo itself is ending up at Gopkg.lock.

sebright:
@ahmetb Thanks for fixing the Go build!  I didn't realize that it wasn't supposed to depend on the Github version.

Did you mean to include 1b61b97be3289c55110213dae7e7b2f95e52042a ?  That is the change I reverted in https://github.com/GoogleCloudPlatform/microservices-demo/pull/60#issuecomment-426119356.

ahmetb:
@sebright ah I just rebased on top of master after merging the #60, so I think they were already here.

Feel free to revert parts of that commit (we'll need the Go parts), but I think it's fine (maybe just doesn't have the `recommendationservice` regenerated).

sebright:
I don't mind if all of the generated file changes are included.  I never pushed the Python code, since it didn't run.

ahmetb:
Build passing, I'll merge and verify.

sebright:
Thanks!

PR #66, opened 2018-10-02
log: change log format to JSON payload for better log in Stackdriver (#47)

ymotongpoo:
change the log format in Python and Node.js services.

Effected services are currencyservice, emailservice, paymentservice,
and recommendationservice. Loadgenerator is left as is because of
the diffculty to change the log format and log target in locust.

### DIFF
--- a/src/emailservice/logger.py
+++ b/src/emailservice/logger.py
@@ -1,0 +1,41 @@
+#!/usr/bin/python
+#
+# Copyright 2018 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import sys
+import json
+
+class JSONStreamHandler(logging.StreamHandler):
+  """JSONStreamHandler emits log data in JSON format to stdout"""
+
+  def __init__(self):
+    super().__init__(sys.stdout)
+
+  def emit(self, record):
+    try:
+      msg = super().format(record)
+      data = {
+        'message': msg,
+        'severity': record.levelname,
+        'name': 'emailservice'
+      }
+      stream = self.stream
+      stream.write(json.dumps(data)+'\n')
+      self.flush()
+    except (KeyboardInterrupt, SystemExit):
+      raise
+    except:
+      self.handleError(record)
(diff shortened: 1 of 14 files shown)


src/emailservice/logger.py
```diff
@@ -0,0 +1,41 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import sys
+import json
+
+class JSONStreamHandler(logging.StreamHandler):
```
ahmetb: I wonder if there's an off-the-shelf library we can use for json logging. 

It would be great to:
- keep the sample short by not writing this implementation
- not duplicate it between two python services

so I tend to think if you can find an external library, it can work better.

ahmetb:
@ymotongpoo thanks a lot. left a comment. I wouldn't bother converting loadgenerator into structured logging, we don't consider it as a "service" to be monitored.

src/emailservice/logger.py
ymotongpoo: Gotcha. [python-json-logger](https://github.com/madzak/python-json-logger) should support it.
I'll replace the code to use this lib.

src/recommendationservice/logger.py
```diff
@@ -0,0 +1,40 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import sys
+from pythonjsonlogger import jsonlogger
+
+# TODO(yoshifumi) this class is duplicated since other Python services are
+# not sharing the modules for logging.
+class CustomJsonFormatter(jsonlogger.JsonFormatter):
```
ahmetb: Is it possible to avoid this file altogether? (since we're still duplicaing it.)
I want to think initializing a SD-compatible logging class shouldn't be 40 lines.

If there's any examples of logging to SD from 12-factor Python apps, I think we should try to reuse those patterns.

ymotongpoo: (Sorry I squashed the commits for better commit logs)

I agree that it would be nice if we can use a neat logging library which enable us to create SD compatible JSON structured logs. I'm still searching the nicer library than python-json-logger but it seems that's the best option so far.

One thing I want to emphasize here is that in Python it's really common to use `logging` standard library and create custom Handler or Formatter for desired log format, even if the format is flat text format, because in Python logging lib, the default format of log levels are in integers and level names are alias of those. In order to extract log level string, we need to have some custom class to extract `levelname` data from `record` object.

Also, the duplication happened due to the project structure. If we can share the module across all Python services, this problems never happens. Moreover, as you say, it's best for SD Logging to provide one-stop library for this use case.

ahmetb: I see. I just started a thread internally to see maybe if there's an utility library we're missing. Truth to be told, this was much easier in Go. Initializing a JSON logger took 2 lines and 5 more lines to specify the custom fields. (We didn't even need to create a new method for it.) Let's wait and see what comes out of internal thread.

PR #67, opened 2018-10-05
Optimized adservice Image Size

### LINKED ISSUE #64: adservice: minify docker image
Currently adservice image is around 900 MB. This probably can be reduced by using an alpine-based image, or moving the compiled artifacts into an alpine-based jre image.

orthros:
The adservice image is now ~165MB (was ~886MB)

Fixes #64 

Uses `openjdk:8-alpine` as a base image and uses multi stage builds for dependency management and building.

## Statistics:

Build time before:
```
time docker build --no-cache -t adservice:master .
...
Successfully tagged adservice:master
real    1m6.333s
user    0m1.543s
sys     0m1.190s
```

Build time after:
```
time docker build --no-cache -t adservice:alpine .
...
Sucessfully tagged adservice:alpine
real    1m14.770s
user    0m1.657s
sys     0m1.262s
```


src/adservice/Dockerfile
```diff
@@ -1,9 +1,17 @@
...
+
+FROM base as builder
+
 WORKDIR /app
 
+# This is because the protoc-gen-grpc-java library 
+# is a glibc compiled binary and alpine is
+# based on musl-libc
+RUN  apk add --no-cache --allow-untrusted \
+    -X https://apkproxy.herokuapp.com/sgerrand/alpine-pkg-glibc \
```
ahmetb: why do we need the untrusted package?

ahmetb: I think we pull the .pub/.apk as shown in the linked repo and avoid --allow-untrusted? That would be a good improvement over current one.

orthros: 👍

src/adservice/Dockerfile
```diff
@@ -1,9 +1,17 @@
...
-    chmod +x /bin/grpc_health_probe
+FROM openjdk:8-alpine as base
+
+FROM base as builder
+
 WORKDIR /app
 
+# This is because the protoc-gen-grpc-java library 
+# is a glibc compiled binary and alpine is
+# based on musl-libc
```
ahmetb: sorry for the back and forth, but have you seen the https://pkgs.alpinelinux.org/package/edge/testing/x86_64/grpc-java

I think I found about it about a month ago and reported to grpc team, it looks like they now documented it https://github.com/grpc/grpc-java/blame/fbfc3a40d073237a3da625e4da300bc7343b26d8/README.md#L157

I'm not sure how to install it, since it's in edge/testing repo.

orthros: Its all good.

I'm unfamiliar with that package. I'll try out and report back

orthros: EDIT: I am preserving this for the sake of knowledge and history. 99% of the way through this comment I had a different idea, which I outline in my next comment.

So I looked at the package and it is easy to install, but only provides a single binary `/usr/bin/protoc-gen-grpc-java` which is not used by the `gradle` build process (as far as I can tell). The actual error is when gradle attempts to generate the protos by calling `protoc-3.5.1-1-linux-x86_64.exe` (compiled as a glibc binary as opposed to a musl-libc binary), which is downloaded, managed and maintained by gradle (its found in the gradle cache).

I attempted to install protoc via `apk` manually in hopes of tricking gradle into using that, but with no such luck. 

I also did a search through GitHub and I don't see the package used at all. 

Big picture, I think using this package is a viable path, but I know I'm probably not the right person to implement it, and I worry we'd be over complicating the Java build process. I think adding a semi obscure compatibility package to our containers for the compile step might be better to improve compatibility across the board

orthros: So I had a thought. Java runs on a VM, and the os architecture doesn't change between steps. 

We could instead do
```dockerfile
FROM openjdk:8-slim

# do the build steps here

FROM openjdk:8-alpine
WORKDIR /app
COPY --from=builder /app .
EXPOSE 9555
ENTRYPOINT ["/app/build/install/hipstershop/bin/AdService"]
```

And we would avoid the glibc incompatibility altogether. In fact, I did this on my machine, and it built and deployed successfully to GKE, and the frontend continued to serve ads.

Thoughts?

ahmetb: Yeah that sounds straightforward. I am not sure if grpc java impl requires libgrpc or something.

PR #78, opened 2018-10-16
Add GCB cloudbuild.yaml config

davidstanke:
This Google Cloud Build (GCB) config deploys to a staging cluster on commit to master.


ahmetb:
I think we already merged this.

Issue #80, opened 2018-10-16
currencyservice - unable to access external endpoint with istio


mukundha:
Currency service uses this endpoint - `http://www.ecb.europa.eu/stats/eurofxref/eurofxref-daily.xml`

Maybe this could be recent change to the endpoint, but it redirect to a https URL, 
so if we enable Istio, we run into the issue described [here](https://istio.io/blog/2018/egress-https/#tls-origination-by-istio)

there are 2 options 
1) Open all egress traffic [diasble istio introspection]
2) maybe find another alternative or host a http equivalent of this endpoint?

Thanks,

ahmetb:
Indeed, the change now poses a problem.

Another good option is to actually hardcode the conversion rates. We only use a few of them anyway. If you have time to refactor the currencyservice to use hardcoded conversion rates, it would be much appreciated.

ahmetb:
cc: @ace-n also, #82.
We might need to hardcode the currency conversion values into the service. This has a potential of breaking live on-stage demos since we don't control it.

ahmetb:
We removed the external dependency, so this should now be obsolete.

PR #82, opened 2018-10-17
Update Currency service to use a http service

mukundha:
Fix for this issue: https://github.com/GoogleCloudPlatform/microservices-demo/issues/80

### DIFF
--- a/istio-manifests/whitelist-egress-currencyprovider.yaml
+++ b/istio-manifests/whitelist-egress-currencyprovider.yaml
@@ -19,6 +19,7 @@ metadata:
 spec:
   hosts:
   - www.ecb.europa.eu
+  - amer-demo4-test.apigee.net
   ports:
   - number: 80
     name: http
--- a/kubernetes-manifests/currencyservice.yaml
+++ b/kubernetes-manifests/currencyservice.yaml
@@ -25,8 +25,7 @@ spec:
       terminationGracePeriodSeconds: 5
       containers:
       - name: server
-        #image: gcr.io/microservices-demo-app/currencyservice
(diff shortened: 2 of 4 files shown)


ahmetb:
We definitely shouldn’t change it to a service that’ll have shorter lifetime than the one we already have.

What’s the rationale for pointing it to your demo endpoint rather than hardcoding the conversion values into the service?

Since people give keynote demos with this app, unfortunately we can’t point it to your demo app.

mukundha:
Good points @ahmetb 
1 -  think would be useful to externalize the currency URL from the code, 
2 - we need a http service, so I proxied the original service using Apigee,  I will probably setup dedicated environment with hipster CNAME -would that work?
3 - i think having a egress rule for this service has value for istio demos and egress policies, hence did not hardcode the conversion values in the service

ahmetb:
> 2 - we need a http service, so I proxied the original service using Apigee, I will probably setup dedicated environment with hipster CNAME -would that work?

I don't think so, unless you're committed to support this service officially for O(N) years.

> 1 - think would be useful to externalize the currency URL from the code,

We literally have 1 external dependency, and it broke in the first 3 months. While I agree, something like this has potentially turn a keynote demo into a complete catastrophe. Since we used this demo for at least 5 keynotes in the past 3 months, I'm hesitant now.

> 3 - i think having a egress rule for this service has value for istio demos and egress policies,

I agree, but this demo app demonstrates much more important aspects of Istio (like Telemetry, mTLS, metrics etc) and not so much of the Egress rules. Unfortunately I don't think we can have it all. It was good while it lasted, now we have a broken demo in our hands.

mukundha:
makes sense, i will close this PR

PR #85, opened 2018-10-23
Toggle bug in ProductCatalogService for demos

### LINKED ISSUE #84: Introducing bug in ProductCatalogService for demos
Hi,

As discussed offline with @ahmetb, I introduced a bug in ProductCatalogService for a keynote demo of Next London 18.

I had some constraints for this demo:
- The bug needed to consume some CPU cycles for the problem to be easily visible in Stackdriver Profiler
- I needed to be able to trigger the bug without having to recreate a pod: in the new Stackdriver IRM product, we can show correlations between an alert and an other metric. It's easier with an existing pod: if you trigger the bug with a new pod, it's harder to get a correlation because all metrics for the pod began exactly at the same time.

Here is what I came up with:
- The bug is just a loop in the `parseCatalog` function that runs for 200 microseconds. This loop consumes some CPU cycles.
- The bug is triggered by a SIGUSR1, and removed (if needed) by SIGUSR2.

I think that this bug can be useful to other people for other demos.
Should I push a PR for that?

MrTrustor:
Fixes #84 

@ahmetb let me know what you think. I've used 150 microseconds because 200 slows the website down too much.

### DIFF
--- a/src/productcatalogservice/README.md
+++ b/src/productcatalogservice/README.md
@@ -3,3 +3,29 @@
 Run the following command to restore dependencies to `vendor/` directory:
 
     dep ensure --vendor-only
+
+## Dynamic catalog reloading / artificial delay
+
+This service has a "dynamic catalog reloading" feature that is purposefully
+not well implemented. The goal of this feature is to allow you to modify the
+`products.json` file and have the changes be picked up without having to
+restart the service.
+
+However, this feature is bugged: the catalog is actually reloaded on each
+request, introducing a noticeable delay in the frontend. This delay will also
+show up in profiling tools: the `parseCatalog` function will take more than 80%
+of the CPU time.
+
+You can trigger this feature (and the delay) by sending a `USR1` signal and
+remove it (if needed) by sending a `USR2` signal:
+
+```
+# Trigger bug
+kubectl exec \
+    $(kubectl get pods -l app=productcatalogservice -o jsonpath='{.items[0].metadata.name}') \
+    -c server -- kill -USR1 1
+# Remove bug
+kubectl exec \
+    $(kubectl get pods -l app=productcatalogservice -o jsonpath='{.items[0].metadata.name}') \
+    -c server -- kill -USR2 1
+```
--- a/src/productcatalogservice/server.go
+++ b/src/productcatalogservice/server.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"os/signal"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -43,9 +44,9 @@ import (
 )
 
 var (
-	catalogJSON []byte
-	cat         pb.ListProductsResponse
-	log         *logrus.Logger
+	cat          pb.ListProductsResponse
+	catalogMutex *sync.Mutex
+	log          *logrus.Logger
 
 	port = flag.Int("port", 3550, "port to listen at")
 
@@ -53,14 +54,6 @@ var (
 )
 
 func init() {
-	c, err := ioutil.ReadFile("products.json")
-	if err != nil {
-		log.Fatalf("failed to open product catalog json file: %v", err)
-	}
-	catalogJSON = c
-	if err := jsonpb.Unmarshal(bytes.NewReader(catalogJSON), &cat); err != nil {
-		log.Warnf("failed to parse the catalog JSON: %v", err)
-	}
 	log = logrus.New()
 	log.Formatter = &logrus.JSONFormatter{
 		FieldMap: logrus.FieldMap{
@@ -71,7 +64,11 @@ func init() {
 		TimestampFormat: time.RFC3339Nano,
 	}
 	log.Out = os.Stdout
-	log.Info("successfully parsed product catalog json")
+	catalogMutex = &sync.Mutex{}
+	err := readCatalogFile(&cat)
+	if err != nil {
+		log.Warnf("could not parse product catalog")
+	}
 }
 
 func main() {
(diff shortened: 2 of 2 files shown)


src/productcatalogservice/server.go
```diff
@@ -148,6 +168,22 @@ type productCatalog struct{}
 func parseCatalog() []*pb.Product {
 	var cat pb.ListProductsResponse
 
+	if problem {
+		done := make(chan int)
```
ahmetb: why not just `time.Sleep()` here? Are you trying to trigger a 100% memory?

If so, please simplify as 

	t := time.Now().Add(150 * time.Millisecond)
	for time.Now().Before(t) {}

src/productcatalogservice/server.go
```diff
@@ -45,6 +47,8 @@ var (
 	log         *logrus.Logger
 
 	port = flag.Int("port", 3550, "port to listen at")
+
+	problem bool = false
```
ahmetb: just `problem bool` is ok

src/productcatalogservice/server.go
MrTrustor: Yes, that's what I'm trying to do. Simplified as suggested.

MrTrustor:
I switched back to using USR2 to remove the bug. I don't have a strong opinion on this, but if you think that `kubectl exec` is not reliable (personally never had an issue with it), it's probably better like this.

src/productcatalogservice/README.md
```diff
@@ -3,3 +3,23 @@
...
+tools.
+
+You can trigger this bug by sending a `USR1` signal and remove it (if needed) by
+sending a `USR2` signal:
+
+```
+# Trigger bug
+kubectl exec \
+    $(kubectl get pods -l app=productcatalogservice -o jsonpath='{.items[0].metadata.name}') \
+    -c server -- kill -USR1 1
```
ahmetb: We're about to add a entrypoint wrapper for Stackdriver Debugger to all these processes in #87 . I am not sure the debugger process is able to forward signals to its child process adequately. Do you mind testing?

MrTrustor: Will do and report back.

ahmetb: Thanks, that'll be great. I am sorry this is very disrupting.

MrTrustor: So, I'm having a hard time figuring out what's happening here:
- The debugger seems to slow down significatively productcatalogservice in any case.
- It also seems to break the logging: I'm getting the first few lines of logs of productcatalogservice, but then nothing, even if I send directly a USR1 or USR2 to the productcatalogservice PID.
- I looked a little bit at the [code of the debugger](https://github.com/GoogleCloudPlatform/google-cloud-go/tree/master/cmd/go-cloud-debug-agent), and it doesn't look like it's forwarding USR1 and USR2. 

We can work around the fact that the debugger doesn't forward the signals, but the slowness introduced and the fact that we don't have logs anymore is a problem.

I'll open a couple of issues in [google-cloud-go](https://github.com/GoogleCloudPlatform/google-cloud-go).

src/productcatalogservice/server.go
```diff
@@ -148,6 +168,12 @@ type productCatalog struct{}
 func parseCatalog() []*pb.Product {
 	var cat pb.ListProductsResponse
 
+	if problem {
+		t := time.Now().Add(150 * time.Microsecond)
+		for time.Now().Before(t) {
+		}
```
MrTrustor: @ahmetb Wait before merging this, I'm getting a feedback that this version of the bug changes how Stackdriver Profiler shows it. I need to retest that.

ahmetb: ACK. I don't know what to do about the debugger issues yet. Let's hold until we figure out a way forward with or without the debugger.

MrTrustor: Ok, so I now better understand what's happening:

- That version of the bug doesn't really slow down the application.
- The previous version (with the `select`) actually slows down significantly the application, even with the same delay, but for a different reason: it consumes a lot more CPU, and with the load-generator running, we're running out of resources.

What do you think? I'd say the simplest is just to go back to that version:
```
	if problem {
		done := make(chan int)
		go func() {
			for {
				select {
				case <-done:
					return
				default:
				}
			}
		}()
		time.Sleep(50 * time.Microsecond)
		close(done)
	}
```

ahmetb: Sounds good. But I still don’t quite understand why we are spinning another goroutine rather than blocking the current one. Feel free to restore the old code.

ahmetb:
We decided not to instrument Go services with debugger just yet because of the issues you pointed out. If the code above is good to you, I'll take this in.

MrTrustor:
@ahmetb I finally took the time to come back to this. I changed my thinking on it.
The delay is now implemented as a broken "reload catalog" feature that actually reads and parse the full catalog on each request.
It introduces a noticeable delay in the frontend and does show up in Stackdriver Profiler.

It still is triggered and removed with USR1/USR2.

Let me know what you think.

src/productcatalogservice/server.go
```diff
@@ -147,13 +171,22 @@ func initProfiling(service, version string) {
...
-		log.Warnf("failed to parse the catalog JSON: %v", err)
-		return nil
+	if reloadCatalog {
+		var catalog pb.ListProductsResponse
+		c, err := ioutil.ReadFile("products.json")
+		if err != nil {
+			log.Fatalf("failed to open product catalog json file: %v", err)
+			return nil
+		}
+		if err := jsonpb.Unmarshal(bytes.NewReader(c), &catalog); err != nil {
```
ahmetb: this is duplicating the code above, including the file name. so it results in less maintainable code

can we have a method named `readCatalogFile()` referenced from both places instead?

ahmetb:
If that gets you what you want, sounds good to me. Left a code comment.

ahmetb:
Looks good. Thanks.

PR #86, opened 2018-10-23
Ignore .vs 

### LINKED ISSUE #83: Remove src/cartservice/.vs
https://github.com/GoogleCloudPlatform/microservices-demo/tree/master/src/cartservice/.vs/cartservice

This directory probably should not be checked in. Contains many binary/cache files. If there's config, we can keep it, but I couldn't find it as such.

orthros:
Fixes #83


.gitignore
```diff
@@ -5,6 +5,7 @@ pkg/
 *.swp
 *~
 .vscode/
+.vs/
 .vs/slnx.sqlite
```
ahmetb: I think you can remove these 2 lines now?
also you might need to do `git rm -rf .vs/`

orthros: My previous commit should have removed the only reference .vs in the codebase `find . -name ".vs"` returns nothing

ahmetb:
@simonz130 are these OK to delete do you think?

simonz130:
NP to remove what you already merged.
Next time give me more time to respond.

ahmetb:
Will do. I quickly double checked to see if .vs is ok to keep out of source control, seemed pretty straightforward.

PR #88, opened 2018-10-24
Add Grafana, Prometheus and Jaeger containers.

rghetia:


rghetia:
Need to install these services through Helm. I'll resubmit the PR with the changes.

PR #93, opened 2018-11-05
adservice: upgrade opencensus-java to 0.17.0.

sebright:
0.17.0 is the first stable version of opencensus-contrib-log-correlation-log4j2.
This commit also updates log4j2.xml to work with the new version.

/cc @rghetia


sebright:
Thanks!

PR #96, opened 2018-11-13
increase reporting interval.

rghetia:
also fixed a bug in SD export registration in Ads.

### DIFF
--- a/src/adservice/src/main/java/hipstershop/AdService.java
+++ b/src/adservice/src/main/java/hipstershop/AdService.java
@@ -207,16 +207,24 @@ public class AdService {
     // Registers logging trace exporter.
     LoggingTraceExporter.register();
     long sleepTime = 10; /* seconds */
-    int maxAttempts = 3;
+    int maxAttempts = 5;
+    boolean statsExporterRegistered = false;
+    boolean traceExporterRegistered = false;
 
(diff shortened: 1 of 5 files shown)


src/adservice/src/main/java/hipstershop/AdService.java
```diff
@@ -207,16 +207,24 @@ public static void initStackdriver() {
     // Registers logging trace exporter.
     LoggingTraceExporter.register();
     long sleepTime = 10; /* seconds */
-    int maxAttempts = 3;
+    int maxAttempts = 5;
+    boolean statsExporterRegistered = false;
+    boolean traceExporterRegistered = false;
```
ahmetb: As far as I can tell you don't use these variables anywhere. Why introduce them?

ahmetb:
FYI @mtwo @rakyll we'll try this patch to see if it brings down the Stackdriver Metrics costs.

src/adservice/src/main/java/hipstershop/AdService.java
ahmetb: oh I see the logic now. I would've preferred if each operation had its own retries like `retry(initTracing())` and `retry(initStats())` but this is OK.

PR #98, opened 2018-11-16
remove gcr repo name from images

### LINKED ISSUE #17: Don't hardcode the GCP project name
The demo hardcodes a GCP project name (microservices-demo-app). Users who don't have access to the hardcoded project cannot run the instructions due to ACL issues.

davidstanke:
This removes hardcoded GCP project name from images and requires an explicit repository flag to skaffold. Also updating the cloudbuild.yaml for staging with the gcr.io/k8s-skaffold/skaffold image.

Fixes #17.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,44 +1,23 @@
 # Cloudbuild.yaml to deploy to staging
 #
 # PREREQUISITES:
-# - Cloud Build service account must have roles: "Kubernetes Engine Developer", "Storage Object Creator", \
-#   "Storage Object"
+# - Cloud Build service account must have roles: "Kubernetes Engine 
+# Developer", "Storage Object Creator", "Storage Object"
+
+# USAGE:
+# GCP zone and GKE target cluster must be specified as substitutions
+# Example invocation:
+# `gcloud builds submit --config=cloudbuild.yaml --substitutions=_ZONE=us-central1-b,_CLUSTER=demo-app-staging .`
 
 steps:
-
-# authenticate to GKE cluster
-- id: 'Get GKE credentials'
-  name: gcr.io/cloud-builders/gcloud
-  args: ['container','clusters','get-credentials','demo-app-staging']
-  env:
-  - 'CLOUDSDK_COMPUTE_ZONE=us-central1-b'
-
-# make latest version of the Skaffold community builder available in this project
-- id: 'clone community builders repo'
-  name: gcr.io/cloud-builders/git
-  args: ['clone','https://github.com/GoogleCloudPlatform/cloud-builders-community']
-
-- id: 'build Skaffold builder'
-  name: gcr.io/cloud-builders/docker
+- id: 'Deploy to app'
+  name: 'gcr.io/k8s-skaffold/skaffold'
+  entrypoint: 'bash'
   args: 
-    [
-      'build',
-      '--tag', 'skaffold',
-      '/workspace/cloud-builders-community/skaffold/.',
-    ]
-
-# push app code to cluster
-- id: 'Deploy to staging'
-  name: skaffold
-  args: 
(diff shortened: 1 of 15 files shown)


cloudbuild.yaml
```diff
@@ -1,17 +1,48 @@
...
 steps:
-- name: gcr.io/k8s-skaffold/skaffold:v0.16.0
-  args: ['skaffold', 'run', '-f=skaffold.yaml']
+# modify skaffold config to deploy to cluster in *current* GCP project
+- id: 'Patch Skaffold config'
+  name: ubuntu
+  args:
+    - 'sed'
+    - '-i'
+    - 's/microservices-demo-app/$PROJECT_ID/g'
```
ahmetb: I think skaffold added a feature for this. I will check next week.

ahmetb:
Thanks for removing sed stuff. Let me know once this is good to go @davidstanke

davidstanke:
Hey @ahmetb thanks for finding that skaffold option. I've cleaned this up a bit and confirmed that it works in my testing project. Please merge!

cloudbuild.yaml
```diff
@@ -1,17 +1,42 @@
...
+      '--tag', 'gcr.io/$PROJECT_ID/skaffold',
+      '/workspace/cloud-builders-community/skaffold/.',
+    ]
+
+# push app code to cluster
+- id: 'Deploy to staging'
+  name: gcr.io/$PROJECT_ID/skaffold
+  args: ['run', '-f=skaffold.yaml']
+  env:
+  - 'SKAFFOLD_DEFAULT_REPO=gcr.io/hipstershop-cicd'
```
ahmetb: is this our default repo?

cloudbuild.yaml
```diff
@@ -1,17 +1,42 @@
...
+  args: ['container','clusters','get-credentials','demo-app-staging']
   env:
   - 'CLOUDSDK_COMPUTE_ZONE=us-central1-b'
+
+# make latest version of the Skaffold community builder available in this project
+- id: 'clone community builders repo'
+  name: gcr.io/cloud-builders/git
+  args: ['clone','https://github.com/GoogleCloudPlatform/cloud-builders-community']
+
+- id: 'build Skaffold builder'
```
ahmetb: why not stick with gcr.io/k8s-skaffold/skaffold image?

cloudbuild.yaml
```diff
@@ -1,17 +1,42 @@
...
+  name: gcr.io/cloud-builders/docker
+  args: 
+    [
+      'build',
+      '--tag', 'gcr.io/$PROJECT_ID/skaffold',
```
ahmetb: you can call this skaffold for simplicity, we aren't pushing this image.

cloudbuild.yaml
davidstanke: I ran into a problem with that... because it's not intended for use as a Cloud Build builder, it failed when used inside GCB (I'm afraid I don't recall what the difference was, or how it failed -- that was pre-Thanksgiving).

cloudbuild.yaml
davidstanke: d'oh! no, that's hardcoded from my test project. fixed in followup commit.

cloudbuild.yaml
davidstanke: good call; changed.

cloudbuild.yaml
ahmetb: It might be because I keep changing the skaffold.yaml spec version and updating skaffold version in .travis.yml.

I strongly recommend using the off-the-shelf image. it saves us 2 steps == better sample readability.

davidstanke:
@ahmetb : updated per our chat today. This version worked on my test project.

PR #103, opened 2018-12-11
Re-enable opencensus python

c24t:
Enables tracing in the email and recommendation services, which was disabled in 316db88 because of a memory leak in the stackdriver exporter.

We fixed the leak in https://github.com/googleapis/google-cloud-python/pull/6856. The fix is included in the [0.1.10 release of opencensus-python](https://github.com/census-instrumentation/opencensus-python/releases/tag/v0.1.10).

With this diff, traces show up as expected in stackdriver while running the demo on GKE. Using an `opencensus-python` package version before `0.1.10` causes the email and recommendation services to leak memory until they OOM. Memory use is back to normal (i.e. roughly constant) using the new package version.


kubernetes-manifests/emailservice.yaml
```diff
@@ -24,7 +24,7 @@ spec:
     spec:
       terminationGracePeriodSeconds: 5
       containers:
-      - name: server
+      - name: emailservice
```
ahmetb: is there a specific reason to do this change? if not, please revert for the sake of consistency with other manifests.

c24t: To show stats aggregated by container name in stackdriver to repro the memory leak. I don't mind changing these back, but a better solution might be to rename all the other containers unless there's a good reason to call them all "server".

screen shot 2018-12-10 at 11 17 17 pm

c24t: Let me know if a33ff61 works for you, if not I'll change them back.

ahmetb: Thanks this LGTM.

ahmetb: Sorry I meant to say: can we revert all services back to name "server"? :)

Ideally you _really_ should not be inferring the "app name" from "container name". Container name can be just "frontend" or "web" or "server", but doesn't have to actually tell what the app does.

Container name is useful only within the context of a Pod and not across the fleet.

ahmetb:
LGTM

PR #104, opened 2018-12-18
Minor update to README.md

charlesbaer:
Added the --region to gcloud container clusters create


ahmetb:
`--region` will create regional clusters (3x many nodes as specified). This won't work for many accounts with 8.0 default cores quota. So please use `--zone`.

charlesbaer:
Updated to --zone

PR #106, opened 2018-12-20
Instrumenting CartService with OpenCensus Trace API

simonz130:
1) Instrumented cart service with OpenCensus Trace API, exporting traces to Stackdriver
2) Migrated cart service+tests to .NET Core 2.1. This required to update base image to install openssl package


src/cartservice/Program.cs
```diff
@@ -41,6 +46,9 @@ class ServerOptions
 
             [Option('r', "redis", HelpText = "The ip of redis cache")]
             public string Redis { get; set; }
+
+            [Option("projectId", HelpText = "The ProjectId to which telemetry will flow")]
```
ahmetb: I'd call this string

    "(optional) GCP project ID to send Stackdriver telemetry data"

src/cartservice/Program.cs
```diff
@@ -104,55 +112,27 @@ static void Main(string[] args)
...
+                            int port = ReadParameter("cart service port", options.Port, CART_SERVICE_PORT, int.Parse, 8080);
 
-                            // Set redis cache host (hostname+port)
-                            ICartStore cartStore;
-                            string redis = ReadRedisAddress(options.Redis);
+                            string projectId = ReadParameter("cloud service project id", options.ProjectId, PROJECT_ID, p => p, null);
 
-                            // Redis was specified via command line or environment variable
-                            if (!string.IsNullOrEmpty(redis))
+                            // Initialize Stackdriver Exporter - currently for tracing only
```
ahmetb: The comment seems unnecessary, and no need to say "currently for x only" as it'll probably change and we'll forget to update this. Let the code speak for what's happening.

src/cartservice/Program.cs
```diff
@@ -104,55 +112,27 @@ static void Main(string[] args)
...
-                                cartStore = new LocalCartStore();
+                                var exporter = new StackdriverExporter(
+                                    projectId, 
+                                    Tracing.ExportComponent,
+                                    viewManager: null);
+                                exporter.Start();
                             }
 
+                            // Set redis cache host (hostname+port)
+                            string redis = ReadParameter("redis cache address", options.Redis, REDIS_ADDRESS, p => p, null);
```
ahmetb: why don't you put all the ReadParameter stuff next to each other (initialization) ––and then start connecting to things like stackdriver or redis?

src/cartservice/Program.cs
```diff
@@ -163,21 +143,48 @@ static void Main(string[] args)
             }
         }
 
-        private static string ReadRedisAddress(string address)
+        /// <summary>
+        /// Reads parameter in the right order
```
ahmetb: unclear what "right order" means.

Maybe just say 

> Reads configuration from the environment, either through command-line flag, or if not specified from the environment variable, and falls back to the default value if neither are specified.

src/cartservice/Program.cs
```diff
@@ -163,21 +143,48 @@ static void Main(string[] args)
...
+            // Environment variable
+            Console.Write($"Reading {description} from environment variable {environmentVariableName}. ");
+            string envValue = Environment.GetEnvironmentVariable(environmentVariableName);
+            if (!string.IsNullOrEmpty(envValue))
             {
-                return redis;
+                try
+                {
+                    var envTyped = environmentParser(envValue);
+                    Console.WriteLine("Done!");
```
ahmetb: are these far too verbose?

can't we just say

`Using env var ${envVariableName}`
`Using cmdline flag ${flagVariableName}`

in the docs?

All the "trying", "done" logs seem a bit excessive.

src/cartservice/Program.cs
```diff
@@ -163,21 +143,48 @@ static void Main(string[] args)
...
+                    return envTyped;
+                }
+                catch (Exception)
+                {
+                    // We assign the default value later on
```
ahmetb: if we're gonna have so many logs for debugging CLI/env parsing, definitely log this exception message, otherwise there's no easy way of telling without modifying code or using debugger.

src/cartservice/Program.cs
```diff
@@ -163,21 +143,48 @@ static void Main(string[] args)
...
+                }
             }
 
-            return null;
+            Console.WriteLine($"Environment variable {environmentVariableName} was not set. Setting {description} to {defaultValue}");
```
ahmetb: "setting" is not the right action here, we're not modifying anything.
use "returning".

src/cartservice/cartstore/InstrumentedCartStore.cs
```diff
@@ -0,0 +1,116 @@
...
+using Hipstershop;
+using OpenCensus.Trace;
+using OpenCensus.Trace.Sampler;
+
+namespace cartservice.cartstore
+{
+    /// <summary>
+    /// Wrapper for Cart Store - instrumented with OpenCensus tracing
+    /// </summary>
+    internal class InstrumentedCartStore : ICartStore
```
ahmetb: Why do we have a wrapper? Can't we just instrument the code? 

Is this a pattern in C#?

src/cartservice/cartstore/InstrumentedCartStore.cs
```diff
@@ -0,0 +1,116 @@
...
+        private readonly ICartStore cartStore;
+        private static ITracer tracer = Tracing.Tracer;
+        private ISpanBuilder initializeSpanBuilder, addItemSpanBuilder, emptySpanBuilder, getItemSpanBuilder, pingSpanBuilder; 
+
+        public InstrumentedCartStore(ICartStore cartStore)
+        {
+            this.cartStore = cartStore;
+
+            // Create Span Builders for tracing
+            initializeSpanBuilder = CreateSpanBuilder("Initialize Cart Store");
```
ahmetb: can't we infer span names from RPCs automatically? (all other services do that) If so, please stick to RPC names for the sake of consistency/searchability.

we shouldn't be guessing what's the span name for things.

also, most other languages don't create a span –spans are created by gRPC OC wrapper automatically. Don't we have that in C# yet?

src/cartservice/cartstore/InstrumentedCartStore.cs
```diff
@@ -0,0 +1,116 @@
...
+                return cartStore.Ping();
+            }
+        }
+
+        public static ICartStore Create(string redis)
+        {
+            ICartStore cartStore;
+
+            // Redis was specified
+            if (!string.IsNullOrEmpty(redis))
```
ahmetb: I think we should perform this `if` outside `ICartStore`.

This sounds a bit like code smell to me. Doing

```
if redis == ""
  return new LocalStore()
else
  return new RedisStore(redis)
```

is way better than

```
interface Store {

  method Create(redis) {
    if redis == ""
      return new LocalStore()
    else
      return new RedisStore(redis)
  }
}
```

because you just made the CartStore interface aware of its storage method (i.e. redis).

So please revert this logic to what it was before (first snippet above).

src/cartservice/Program.cs
ahmetb: Right now no other service requires `projectID` parameter/variable.

They all have OC SD Exporter libraries that are capable of discovering whether it's running on GCE, and querying projectID from the GCE metadata service.

src/cartservice/Program.cs
simonz130: It helps to troubleshoot what is happening and it happens only at the startup with 3 parameters. It's very useful to know what happens at the startup when something goes wrong. I'm not going to change it.

src/cartservice/Program.cs
simonz130: In the eyes of the person that is looking at the traces, we are setting the value to default. The comment is right.

src/cartservice/Program.cs
simonz130: There is already a log message that says which source assigns the value to the parameter, so if it's not environment variable, we'll discover it from the logs. No need to add anything in exception handling logic.

src/cartservice/cartstore/InstrumentedCartStore.cs
simonz130: Spans should be named after the logical operation they represent and not after a method name. Method names are implementation detail that can change as a result of refactoring. Span names should never change as it will bust correlation between current and historical data about the same operation.

src/cartservice/cartstore/InstrumentedCartStore.cs
simonz130: We don't have grpc native instrumentation in C# yet. I am planning to add that in the next 1.5 months. Even if we would have had that, it only automatically instruments remote procedure calls. If we want to instrument logical operations in cart service itself, or calls to third party libraries that are not natively instrumented by opencensus, we must instrument ourselves. Other services in Hipstershop don't work with external services so much, so they might not need this.

src/cartservice/Program.cs
```diff
@@ -163,21 +143,48 @@ static void Main(string[] args)
...
             {
-                return address;
+                return commandLineValue;
             }
 
-            Console.WriteLine($"Reading redis cache address from environment variable {REDIS_ADDRESS}");
-            string redis = Environment.GetEnvironmentVariable(REDIS_ADDRESS);
-            if (!string.IsNullOrEmpty(redis))
+            // Environment variable
+            Console.Write($"Reading {description} from environment variable {environmentVariableName}. ");
```
SergeyKanzhelev: Perhaps use ILogger instead of Console here?

simonz130: Do you mean ILogger from Microsoft.Extensions.Logging nuget? 
In this sample we don't use ASP.NET - cart microservice uses grpc (and grpc host) for communication and overall design tries to follow 12 factor apps design, where logs should go to std.out, so sidecars can read them and transparently transfer them out. So we don't use Microsoft.Extensions.* packages or any other dependencies that are not absolutely required.

SergeyKanzhelev: `ILogger` is easy configurable for console or debug output. Moreover many libraries already use ILogger. So I'd encourage to use it here.

``` csharp
serviceCollection.AddSingleton(new LoggerFactory()
    .AddConsole()
    .AddDebug());
```

simonz130: This is true for ASP.NET Core with built-in dependency injection. Grpc C# library doesn't support dependency injection and so we don't have it here too.

src/cartservice/cartstore/InstrumentedCartStore.cs
ahmetb: I guess it's fine and still works the same way, so no need to change.

src/cartservice/cartstore/InstrumentedCartStore.cs
ahmetb: I think span titles for internal operations or external calls are fine, but I assume we should hold off on to this PR until C# can use HTTP request paths from GRPC as span names?

I'm happy to merge this if you say so, but you'll need to refactor once it's available.

DanSanche:

Issue #109, opened 2018-12-27
Export traces to Jaeger if installed


ahmetb:
(This came up in #108 and I believe some work is already happening.)

I'd love to have the demo app export traces to a local Jaeger installation ideally with minimal to no config changes.

I'm not sure how we can detect a jaeger installation uniformly and register it as an exporter (that works alongside with Stackdriver exporter) in multiple languages.

cc: @rghetia

rghetia:
@ahmethb Jaeger itself doesn't support any discovery mechanism. Is environment variable approach not sufficient? Alternatively, a simple GET query on port 16686 of jaeger service (assuming the service is named jaeger) can be used to determine if the service is available or not.

btw, I haven't got a chance to look at adding jaeger/grafana/promethues  to microservice demo. If some can help add those service then I can add the instrumentation part.

ahmetb:
> Jaeger itself doesn't support any discovery mechanism. Is environment variable approach not sufficient?

how would this look like? 

Right now we have a simple check like `if isOnGCE(){registerStackdriverExporter()}` in all languages. (Some languages have this explicitly, and some languages have it in the Stackdriver SDK register method, which fails silently if it's not on GCE).

Also would it be possible to have multiple exporters (i.e. both jaeger and gce) simultaneously?

rghetia:
yes, it is possible to have multiple exporter.
In each service specific .yaml file set env (e.g JAEGER_ENABLE=true). This does require user to change the yaml file though. Not sure if there is a way to set the env for all micro-services while deploying it (or using some global config).

ahmetb:
We can take a shortcut like `helm init && helm install stable/jaeger-operator` to install [this Helm chart](https://github.com/helm/charts/tree/master/stable/jaeger-operator) in the README if folks want to try Jaeger integration.

Ideally we shouldn't maintain Jaeger kubernetes deployment manifests.

ahmetb:
That looks like a cleaner solution. Is the oc-agent ready for use here? If so, I'd be happy to scrub SD/Jaeger specific code from each microservice.

rghetia:
OcAgent is ready. 
Here is how it is used in our [interoptest](https://github.com/census-ecosystem/opencensus-experiments/tree/master/interoptest). It is based on hipstershop-demo model.
 
OcAgent export for
[...]

PR #111, opened 2019-01-03
Add Jaeger support for Adservice.

rghetia:
This is the first service that exports to jaeger. Others to follow.
Requires jaeger to be instantiated using 

- helm install --name jaeger stable/jaeger-operator
- kubectl apply -f jaeger.yaml

=== jaeger.yaml Content ===
apiVersion: io.jaegertracing/v1alpha1
kind: Jaeger
metadata:
  name: jaeger

Above steps will be added to README in subsequent PR.

### DIFF
--- a/kubernetes-manifests/adservice.yaml
+++ b/kubernetes-manifests/adservice.yaml
@@ -31,6 +31,10 @@ spec:
         env:
         - name: PORT
           value: "9555"
+        - name: JAEGER_ENABLED
+          value: "false"
+        - name: PROMETHEUS_ENABLED
+          value: "false"
         resources:
           requests:
             cpu: 200m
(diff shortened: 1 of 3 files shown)


rghetia:
Partially fixes #109.
@ahmetb please take a look.

ahmetb:
I opened this issue to track the work for adding Jaeger support parity to other services: https://github.com/GoogleCloudPlatform/microservices-demo/issues/112

I think we shouldn't do Prometheus in this PR (nobody asked for it yet?) because I feel like it'll take us some time to even get the parity on other services wrt Jaeger. What do you think?

ahmetb:
> With Prometheus and Grafana support it is easy to check OC metrics locally.

yes but we won't be documenting how to install Prometheus or Grafana yet. We'll just update readme on how to install+enable Jaeger.

That's why I was hesitant about getting code that won't be used in.

rghetia:
Let me remove Prometheus for now.

ahmetb:
We should start documenting this after maybe we have 2-3 more services that support jaeger? Do you have more services working with Jaeger in your fork?

PR #113, opened 2019-01-05
Add jaeger support for frontend.

rghetia:


rghetia:
@ahmetb PTAL.

src/frontend/main.go
```diff
@@ -158,16 +182,26 @@ func initStats(log logrus.FieldLogger, exporter *stackdriver.Exporter) {
 }
 
 func initTracing(log logrus.FieldLogger) {
+	// This is a demo app with low QPS. trace.AlwaysSample() is used here
```
ahmetb: Hmm I recall we had this already in. I wonder at what point it got removed. Thanks for adding back.

src/frontend/main.go
```diff
@@ -142,6 +143,29 @@ func main() {
 	log.Fatal(http.ListenAndServe(addr+":"+srvPort, handler))
 }
 
+func initJaegerTracing(log logrus.FieldLogger) {
+
+	svcAddr := os.Getenv("JAEGER_SERVICE_ADDR")
```
ahmetb: This env var doesn't match what we had in the Java adservice. It was a bool; here it's an addr with !="" check.

I think we should unify the discovery?

rghetia: yes, I plan to modify the jave to match this one. I like this approach because it allows the user to name the endpoints.

ahmetb: agreed this is better.

PR #116, opened 2019-01-09
Add jaeger export to checkoutservice.

rghetia:


src/checkoutservice/main.go
```diff
@@ -109,6 +132,14 @@ func initStats(exporter *stackdriver.Exporter) {
 }
 
 func initTracing() {
+	// This is a demo app with low QPS. trace.AlwaysSample() is used here
+	// to make sure traces are available for observation and analysis.
+	// In a production environment or high QPS setup please use
+	// trace.ProbabilitySampler set at the desired probability.
+	trace.ApplyConfig(trace.Config{DefaultSampler: trace.AlwaysSample()})
```
ahmetb: Do we have to set this on all services? I assumed sample=y/n is enough to set on the frontend, which all traces to this service originate from.

src/checkoutservice/main.go
```diff
@@ -98,6 +99,28 @@ func main() {
...
+		log.Info("jaeger initialization disabled.")
+		return
+	}
+
+	// Register the Jaeger exporter to be able to retrieve
+	// the collected spans.
+	exporter, err := jaeger.NewExporter(jaeger.Options{
+		Endpoint: fmt.Sprintf("http://%s", svcAddr),
+		Process: jaeger.Process{
+			ServiceName: "checkoutservice",
```
ahmetb: Oh I wonder how does stackdriver exporter automatically know the name of the service (I don't think we pass it below?) but jaeger doesn't.

No action needed for this, just chatting.

src/checkoutservice/main.go
```diff
@@ -109,6 +132,14 @@ func initStats(exporter *stackdriver.Exporter) {
...
 
 func initTracing() {
+	// This is a demo app with low QPS. trace.AlwaysSample() is used here
+	// to make sure traces are available for observation and analysis.
+	// In a production environment or high QPS setup please use
+	// trace.ProbabilitySampler set at the desired probability.
+	trace.ApplyConfig(trace.Config{DefaultSampler: trace.AlwaysSample()})
+
+	initJaegerTracing()
+
```
ahmetb: i think we need to move rest of this method to `initStackdriverTracing()` and call it from here. what do you think?

(ditto for all go services, sorry for not caching this earlier)

PR #120, opened 2019-01-11
Simplify Python services

### LINKED ISSUE #58: image build time doubled after image size optimizations
Below is a snapshot of the build times listed at https://travis-ci.com/GoogleCloudPlatform/microservices-demo/builds.

It looks like after merging PRs

1. #51 emailservice image optimization
2. #52 loadgenerator image optimizations

the build times has risen from 12 minutes on average to 22 minutes (after #51), and then to **26 minutes** (after #52).

So we're probably providing a faster image push experience, but now the overall build time has doubled. I assume we're probably now serving the developers with high upload bandwidths worse.

image

cc: @orthros

di:
This PR does a few things:

1. **Removes unnecessary Python dependencies currently being installed for `emailservice`**

    There are quite a few packages being installed that aren't actual dependencies.

2. **Removes a number of related, also unnecessary system-level dependencies for `emailservice`**

    These were a result of the Python dependencies that are unnecessary.

3. **Pins all of the sub-dependencies for `loadgenerator`**

    This is good practice to ensure that things don't break one day in the future when a newer version of an unpinned sub-dependenency is released.

4. **Compile all Python dependencies from `requirements.in` files**

    This is mostly bookkeeping. It allows us to only specify the top-level dependencies we care about in the requirements.in files, which are then compiled to frozen dependencies in the requirements.txt files. This ensures that we only install the dependencies we need, and that we're not missing any unpinned sub-dependencies. It also makes it more clear where our sub-dependencies are coming from.

5. **Switch to -slim images from -alpine**

    Python's built distribution format (wheel) is incompatible with alpine-based images, causing dependencies like `grpcio` to be compiled from scratch, rather than from a pre-built wheel.

    This should improve or possibly fi​x #58, while keeping the image size roughly the same:

    ```
    emailservice          latest           d1b818eabe05        6 seconds ago       286MB
    loadgenerator         latest           4d9b5acbfbbb        6 seconds ago       125MB
    ```


ahmetb:
@di This is looking great. 👍 It definitely has made `emailservice` build a few times faster.

We opted in to use alpine base for both build/distribution of that image, so by using musl libc, we were ending up recompiling all native dependencies like grpc, which tool a long time.

As you said we were blindly doing `pip freeze > requirements.txt`.

ahmetb:
This has cut the build times from 25 minutes to 11 minutes (what we used to have prior to image size improvements).
https://travis-ci.com/GoogleCloudPlatform/microservices-demo/pull_requests

@orthros maybe we should close #58.

orthros:
Agreed. :+1: 
LGTM

PR #124, opened 2019-01-14
On deploy from Cloud Build, stamp the frontend with source hash

davidstanke:
Hey Ahmet, we discussed this a while back... it will help with CICD if we can easily trace the currently-deployed version back to its source. This is a small script that stamps the git hash into the frontend, by replacing a comment in the footer HTML source.

(For local development, this has no effect -- the comment simply remains as a comment.)

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -3,7 +3,7 @@
 #
 # PREREQUISITES:
 # - Cloud Build service account must have role: "Kubernetes Engine Developer"
-
+#
 # USAGE:
 # GCP zone and GKE target cluster must be specified as substitutions
 # Example invocation:
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -23,11 +23,7 @@ steps:
 
 steps:
 - id: 'Deploy application to cluster'
-<<<<<<< HEAD
   name: 'gcr.io/k8s-skaffold/skaffold:v0.20.0'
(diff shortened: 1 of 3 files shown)


ahmetb:
This is a bit hacky and only works as GCB-specific. Since we don't need it right away, I think I'm currently -1.

I would be supportive of this if we can find a way to use `ARG` in Dockerfile so it can be tied back to skaffold somehow.

Ideally we should stamp all binaries with git commit, but I am not sure if skaffold can do that (passing gitCommit as docker build arg) so it might be a feature request.

davidstanke:
@ahmetb Well, it can be easily expanded to support the git commit variables in any CI system, but I can see your point that it would be more elegant to do it as a Docker build arg. I'll play with that, and request a Skaffold feature if needed.

ahmetb:
@davidstanke yeah I don't want to rush things. I'm in favor of getting the products/tools fixed rather than unblocking the CI to work perfectly. The whole point of this app is to test and reveal issues with our tooling/products anyway. :)

ahmetb:
I'm tempted to close this for the time being.

I think when we can cleanly do this during the actual build step (i.e. Dockerfile), I'd be down. Otherwise it's not really adding value, but it's adding inconsistency (between Cloud Build vs manual Skaffold deployment) and complexity.

DanSanche:

PR #126, opened 2019-01-20
add jaeger support for productcatalog and shipping.

rghetia:


src/frontend/main.go
```diff
@@ -215,6 +207,18 @@ func initTracing(log logrus.FieldLogger) {
 	log.Warn("could not initialize stackdriver exporter after retrying, giving up")
 }
 
+func initTracing(log logrus.FieldLogger) {
+	// This is a demo app with low QPS. trace.AlwaysSample() is used here
+	// to make sure traces are available for observation and analysis.
+	// In a production environment or high QPS setup please use
+	// trace.ProbabilitySampler set at the desired probability.
+	trace.ApplyConfig(trace.Config{DefaultSampler: trace.AlwaysSample()})
```
ahmetb: again, should we put the sampling rate to the service where requests don't start from?

src/frontend/main.go
```diff
@@ -215,6 +207,18 @@ func initTracing(log logrus.FieldLogger) {
...
+
+	initJaegerTracing(log)
+	initStackDriverTracing(log)
```
ahmetb: typo: Stackdriver

src/frontend/main.go
ahmetb: yeah, requests originate from frontend. so I'm not sure if having this here does anything in particular.

ahmetb: my bad, this is the frontend code I'm reading.

PR #132, opened 2019-01-28
Adds make-release script, static k8s manifests

### LINKED ISSUE #75: Provide tagged releases/images
It should be easy for someone who is just playing with this or deploying this app for a demo to take YAMLs and apply them to the cluster.

This requires:
1. tagging git releases
2. making pre-built container images available on gcr publicly (with :tags)
3. since tutorials etc will rely on this, also need long-term storage for images (e.g. `gcr.io/google-samples`)

m-okeefe:
Closes #75.  

### Changelog 

Adds 4 scripts to the `hack/` directory for building/pushing images, injecting images tags into static manifests, and tagging new releases. See [hack/README.md](https://github.com/m-okeefe/microservices-demo/tree/release-script/hack). 

**Note**: since we have not pushed images yet, the images in the `./release/` manifests are still set to the skaffold defaults (eg. `adservice`).


hack/make-release.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
```
ahmetb: #!/usr/bin/env bash

so we use bash4

hack/make-release.sh
```diff
@@ -0,0 +1,54 @@
...
+
```
ahmetb: add `set -euo pipefail`

hack/make-release.sh
```diff
@@ -0,0 +1,54 @@
...
+# set vars 
+if [ -n "$1" ]; then
```
ahmetb: - use `[[` not `[`

- make sure you document usage of script simply at the beginning of the script.

hack/make-release.sh
```diff
@@ -0,0 +1,54 @@
...
+    DOCKER_REPO=$1
```
ahmetb: make sure to use quotes around variables to prevent splitting `"$1"`

hack/make-release.sh
```diff
@@ -0,0 +1,54 @@
...
+else
+    echo "Must provide a docker repo"
```
ahmetb: 1. log to stderr
2. exit with error code


I recommend you create 2 funcs for logging stuff


log() { echo "$*" >&2 }
fail() { log "$*"; exit 1 }

hack/make-release.sh
```diff
@@ -0,0 +1,54 @@
...
+fi
+
+if [ -n "$2" ]; then
+    TAG=$2
+else
+    echo "Must provide a version tag."
+    exit 
+fi
+
+if [ -n "$3" ]; then
```
ahmetb: this should always be master anyway.

hack/make-release.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# set vars 
+if [ -n "$1" ]; then
+    DOCKER_REPO=$1
+else
+    echo "Must provide a docker repo"
+    exit 
+fi
+
+if [ -n "$2" ]; then
```
ahmetb: instead of relying on positional args, you can simply do:

    TAG="${TAG?TAG env variable must be specified}"

and you don't deal with ifs, positional arg parsing etc.

hack/make-release.sh
```diff
@@ -0,0 +1,54 @@
...
+
+if [ -n "$3" ]; then
+    BRANCH=$3
+else
+    echo "Must provide a git branch."
+    exit 
+fi
+
+# iterate over all services: build + push image, inject tag into k8s manifest. 
+for dir in ./src/*/
```
ahmetb: hmm use `find -type d` and even `-min-depth 1`, `-max-depth 1 for this, so that it doesn't include /src.

hack/make-release.sh
```diff
@@ -0,0 +1,54 @@
...
+else
+    echo "Must provide a git branch."
+    exit 
+fi
+
+# iterate over all services: build + push image, inject tag into k8s manifest. 
+for dir in ./src/*/    
+do
+    # docker build + push 
+    svcname=$(basename $dir)
```
ahmetb: add double quotes around subshells + the $dir

hack/make-release.sh
```diff
@@ -0,0 +1,54 @@
...
+    image="$DOCKER_REPO/$svcname:$TAG"
+    echo "Building and pushing $image..." 
+    docker build -t $image -f $dir/Dockerfile $dir
```
ahmetb: use "" around all args that has a variable in them.

also not sure why you're doing `-f` here.

you can just do

```
(
   cd "$dir"
   docker build -t "$svcname"
)
```

hack/make-release.sh
```diff
@@ -0,0 +1,54 @@
...
+do
+    # docker build + push 
+    svcname=$(basename $dir)
+    image="$DOCKER_REPO/$svcname:$TAG"
+    echo "Building and pushing $image..." 
+    docker build -t $image -f $dir/Dockerfile $dir 
+    docker push $image 
+
+    # inject new tag into the relevant k8s manifest
+    pattern=".*image:.*"
```
ahmetb: this would capture `simage:` too.

instead do `^\s+image`

hack/make-release.sh
```diff
@@ -0,0 +1,54 @@
...
+    replace="        image: $image"
```
ahmetb: this number of spaces look hacky and fragile. I think you need to use groups to replace only matches with parantheses.

hack/make-release.sh
```diff
@@ -0,0 +1,54 @@
...
+    manifestfile="./release/kubernetes-manifests/$svcname.yaml"
+    sed -i '' "s|$pattern|$replace|g" $manifestfile 
+done
+ 
+
+# create + push new release tag 
+echo "Pushing git tag..."
```
ahmetb: let's split pushing git and building images+dist/ into a separate script

hack/make-release.sh
```diff
@@ -0,0 +1,54 @@
...
+git tag $TAG 
+git push --tags
```
ahmetb: I'd personally do this:

```
git add release/
git commit --allow-empty -m "Release v$TAG"
git tag v"$TAG"
git push --follow-tags
```

release/kubernetes-manifests/adservice.yaml
```diff
@@ -0,0 +1,65 @@
+# Copyright 2018 Google LLC
```
ahmetb: lets merge these all these yamls into one deployment.yaml or something

hack/make-release.sh
m-okeefe: Not 100% sure I know what you mean here. Do you mean capture groups? Why parentheses?

PR #133, opened 2019-01-29
hack: polish scripts

ahmetb:


m-okeefe:
Small thing: when I first ran `./make-release-artifacts.sh`, I saw:

```
./hack/make-release-artifacts.sh: line 57: realpath: command not found
Successfully saved merged manifests to .
``` 

Ran `brew install coreutils`, then it was fine: 

```
Successfully saved merged manifests to /Users/mokeefe/ ....
```

Add note that you need `realpath`?

hack/make-docker-images.sh
```diff
@@ -12,28 +14,28 @@
...
+while IFS= read -d $'\0' -r dir; do
+    # build image
+    svcname="$(basename "${dir}")"
+    image="${REPO_PREFIX}/$svcname:$TAG"
+    (
+        cd "${dir}"
+        log "Building: ${image}"
+        docker build -t "${image}" .
+
+        echo "Pushing: ${image}"
```
m-okeefe: s/echo/log ?

hack/make-release-artifacts.sh
```diff
@@ -14,42 +16,42 @@
...
+    while IFS= read -d $'\0' -r file; do
+        cat "${file}"
+        echo "---"
+    done < <(find "${src_manifest_dir}" -name '*.yaml' -type f -print0)
+}
 
-# remove empty lines 
-gsed -r -i '/^\s*$/d' $tmp 
+# read and merge all manifests
+out_manifest="$(read_manifests)"
```
m-okeefe: trying to figure out how you're getting 1 google header to stay? 
also it's 2018 vs. 2019 now?

ahmetb:
> Add note that you need `realpath`?

we can remove `realpath` dependency.

hack/make-release-artifacts.sh
ahmetb: I removed it because the license was there already, in the original files.

PR #134, opened 2019-01-29
productcatalog: introduce configurable latency

m-okeefe:


src/productcatalogservice/server.go
```diff
@@ -143,6 +144,25 @@ func initStats(exporter *stackdriver.Exporter) {
...
+	}
+	v, err := time.ParseDuration(s)
+	if err != nil {
+		panic("Invalid EXTRA_LATENCY var, must be time.Duration")
+	}
+	sleep = v
+	log.Infof("extra latency enabled (duration: %v)", sleep)
+}
+
+func injectSleep() {
```
ahmetb: This is unnecessary.

src/productcatalogservice/server.go
```diff
@@ -143,6 +144,25 @@ func initStats(exporter *stackdriver.Exporter) {
 	}
 }
 
+// sets the sleepSeconds variable, if env variable provided
+func initSleep() {
+	s := ""
+	if s = os.Getenv("EXTRA_LATENCY"); s == "" {
```
ahmetb: You can do all the following in a single if in main(). Don’t introduce this initSleep method. 

you can use if s := syntax also not to do s:=“”.

src/productcatalogservice/server.go
```diff
@@ -48,6 +48,7 @@ var (
 	cat          pb.ListProductsResponse
 	catalogMutex *sync.Mutex
 	log          *logrus.Logger
+	sleep        time.Duration
```
ahmetb: Call this extraLatency

PR #168, opened 2019-03-01
adservice: use Stackdriver special field to set sampling decision in log entries

sebright:
This commit sets the new sampling decision field that is recognized by the
Stackdriver Logging agent, "logging.googleapis.com/traceSampled".  The sampling
decision field was added in
https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/297, and
it won't be available until the new version of fluent-plugin-google-cloud is
used in GKE.

/cc @rghetia

### DIFF
--- a/src/adservice/src/main/resources/log4j2.xml
+++ b/src/adservice/src/main/resources/log4j2.xml
@@ -4,12 +4,15 @@
     <Console name="STDOUT" target="SYSTEM_OUT">
 
       <!-- This is a JSON format that can be read by the Stackdriver Logging agent. The trace ID,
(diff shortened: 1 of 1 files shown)


ahmetb:
@sebright thanks.
It appears like the adservice is drifting away from other services in terms of features like this.

Ideally we should strive to open an issue to track which services have this feature and try to go add this to most of them.

Some new projects like Stackdriver Sandbox are going to be using this demo app to showcase such features, in that case, they _may_ need the feature to be demo-able in more than one programming languages.

ahmetb:
@sebright fyi ^ this comment if you missed.

sebright:
Thanks for the review!  I opened #188 for the other services.

PR #205, opened 2019-04-05
add ocagent and change go services to use ocagent.

rghetia:


ahmetb:
I didn't take a closer look yet, but this sounds like a lot of hard work, thanks for that. Can we hold off this until after Next? I'm not sure if people are properly pinning to a commit in this repo for their demos, and we don't want to break them with development commits during the conference.

rghetia:
> I didn't take a closer look yet, but this sounds like a lot of hard work, thanks for that. Can we hold off this until after Next? I'm not sure if people are properly pinning to a commit in this repo for their demos, and we don't want to break them with development commits during the conference.

@mtwo wanted this change for Next. He can probably use my fork.

ahmetb:
Yes, this looks like it's on the right track, we should probably come back to it after Next just to be on the safe side.

kubernetes-manifests/ocagent.yaml
```diff
@@ -0,0 +1,66 @@
...
+data:
+  oc-agent-config: |
+    receivers:
+      opencensus:
+        address: "0.0.0.0:55678"
+      jaeger: {}
+      zipkin: {}
+    exporters:
+      stackdriver:
+        # TODO: find a different way to specify project-id
```
ahmetb: can we make oc-agent determine the project-id from the metadata server, just like how all stackdriver services do?

Ideally we should not have a projectID configuration in yaml. Currently the demo app is very easy to deploy because we don't require users to change anything, and we should try to keep it that way.

ahmetb:
This mostly looks good to me. I have a concern around the requirement of configuring oc-agent with the GCP project ID, ideally this should be picked up from the environment (GCE metadata API).

It's gonna be a little weird to have some libraries use OC SD exporter and some use OC Agent exporter (until we port the rest). Do you foresee any issues about this?

mtwo:
> It's gonna be a little weird to have some libraries use OC SD exporter and some use OC Agent exporter (until we port the rest). Do you foresee any issues about this?

@rghetia I suggest that we have all of the libraries pass telemetry through the agent rather than just some. Thoughts?

ahmetb:
@mtwo I think we'll eventually get there, it will be better to do it in small PRs (like this one). I was just curious if there's gonna be a problem while some traces are going through direct SD exporter library, and some going through OC Agent exporter library.

kubernetes-manifests/ocagent.yaml
rghetia: I had same concern. It should be possible but let me check with @songy23.

rghetia:
> @mtwo I think we'll eventually get there, it will be better to do it in small PRs (like this one). I was just curious if there's gonna be a problem while some traces are going through direct SD exporter library, and some going through OC Agent exporter library.

Exporting via agent and directly to SD works for the most part except for the resources. Resources gets dropped via agent. This is again a concern. I have a TODO item to check with.
I do have other services (python and java) done as well. The only remaining services are nodejs and c#. I broke it down by language so it is easy to review.

rghetia:
> This mostly looks good to me. I have a concern around the requirement of configuring oc-agent with the GCP project ID, ideally this should be picked up from the environment (GCE metadata API).
> 
> It's gonna be a little weird to have some libraries use OC SD exporter and some use OC Agent exporter (until we port the rest). Do you foresee any issues about this?

There is [PR](https://github.com/census-instrumentation/opencensus-service/pull/526) open get the projectID from environment when deployed in GCP. I'll wait for the fix.

kubernetes-manifests/ocagent.yaml
songy23: https://github.com/census-instrumentation/opencensus-service/pull/526 should fix it, you can leave this field blank and Agent will use the default project ID.

songy23:
FYI OC-Service 0.1.6 has been released: https://github.com/census-instrumentation/opencensus-service/releases/tag/0.1.6, which includes the fix to default GCP project ID.

rghetia:
> FYI OC-Service 0.1.6 has been released: https://github.com/census-instrumentation/opencensus-service/releases/tag/0.1.6, which includes the fix to default GCP project ID.

thanks @songy23 , I'll update this PR to use released OC-Agent.

rghetia:
updated with new Ocagent (ver 0.1.6). With this version project ID for stackdriver export is not required.

rghetia:
Keeping this on hold until other services work. As of now,
1. python is working but not included in this commit.
2. Java is sort of working. Agent receives the export but unable to export to SD. (Need to investigate).
3. Node is not instrumented yet.

With Istio it doesn't work. So far only tried with Go and Python but it doesn't work with either.

ahmetb:
> With Istio it doesn't work. So far only tried with Go and Python but it doesn't work with either.

@rghetia Is it because DaemonSets are not working properly with Istio? If that's the case, we may figure something out.

rghetia:
> > With Istio it doesn't work. So far only tried with Go and Python but it doesn't work with either.
> 
> @rghetia Is it because DaemonSets are not working properly with Istio? If that's the case, we may figure something out, such as dropping the sidecar on DaemonSet pods etc.

I haven't looked into detail but it seems like services are not able to export to OcAgent running in DaemonSet. if you let me know how to drop the sidecar for DaemonSet then I can try it out.

varungbt:
> > > With Istio it doesn't work. So far only tried with Go and Python but it doesn't work with either.
> > 
> > 
> > @rghetia Is it because DaemonSets are not working properly with Istio? If that's the case, we may figure something out, such as dropping the sidecar on DaemonSet pods etc.
> 
> I haven't looked into detail but it seems like services are not able to export to OcAgent running in DaemonSet. if you let me know how to drop the sidecar for DaemonSet then I can try it out.

I see the same issue. When I use istio, I am unable to send over traces to opencensus collector too.

ahmetb:
@rghetia I see you pushed some commits on Sep 6, is this still WIP or should we try to merge?

rghetia:
> @rghetia I see you pushed some commits on Sep 6, is this still WIP or should we try to merge?

Let me resolve merge conflict first. The other thing is that the new version of OcAgent is OtelSvc from deferent code base. Same functionality but the OcAgent is now in maintenance mode. I was thinking to migrate to OtelSvc before merging it.

DanSanche:

PR #207, opened 2019-04-09
Follow OWASP best practices on CC presentation on checkout form

chroto:
It seems innocuous for a demo like this but unmasked CVV and CC number fields can raise a few eyebrows from PCI-minded folks. I think it's best just to take care of it

See OWASP recommendations here:
<https://www.owasp.org/index.php/Handling_E-Commerce_Payments#Displaying_portions_of_the_credit_card>

### DIFF
--- a/src/frontend/templates/cart.html
+++ b/src/frontend/templates/cart.html
@@ -95,8 +95,7 @@
                                 <div class="form-row">
                                     <div class="col-md-6 mb-3">
                                         <label for="credit_card_number">Credit Card Number</label>
-                                        <input type="password" class="form-control" id="credit_card_number"
-                                            autocomplete="off"
+                                        <input type="text" class="form-control" id="credit_card_number"
                                             name="credit_card_number"
                                             placeholder="0000-0000-0000-0000"
                                             value="4432-8015-6152-0454"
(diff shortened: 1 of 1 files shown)


ahmetb:
I think this makes the demo hard to understand.
I also simply have never seen masked credit card number field (maybe CVV is ok), you can check out payment systems like Stripe, I don't think any of them mask the CC field.

chroto:
@googlebot recheck

chroto:
googlebot recheck

chroto:
Fixed the OWASP link in the description: https://www.owasp.org/index.php/Handling_E-Commerce_Payments#Displaying_portions_of_the_credit_card

chroto:
> I think this makes the demo hard to understand as we pre-populate a box that will look like ****************.
> I also simply have never seen masked credit card number field (maybe CVV is ok), you can check out payment systems like Stripe, I don't think any of them mask the CC field.

FWIW I agree. I've seen little consistency on this subject out in the wild. Stripe's [product page for their Checkout](https://stripe.com/payments/checkout) speaks to this. They show plenty of samples on that page with all combinations of input masking for CC number and CVV. 

Obviously, it's your call to make. If it helps, I can supply more context on the specific use-case that prompted this PR

ahmetb:
Let’s do it only for CVV?

chroto:
> Let’s do it only for CVV?

Sounds good. The input type for the credit card number is reverted back to `type="text"`

PR #209, opened 2019-04-23
Minikube installation instructions added

idvoretskyi:
Tested with Minikube+KVM (4 CPU; 4GB RAM).

Signed-off-by: Ihor Dvoretskyi <ihor@linux.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,11 +68,16 @@ Find **Protocol Buffers Descriptions** at the [`./pb` directory](./pb).
 
 ## Installation
 
-We offer three installation methods:
+We offer the following installation methods:
 
-1. **Running locally with “Docker for Desktop”** (~20 minutes) You will build
+1. **Running locally** (~20 minutes) You will build
    and deploy microservices images to a single-node Kubernetes cluster running
-   on your development machine.
+   on your development machine. There are two options to run a Kubernetes
+   cluster locally for this demo:
+   - [Minikube](https://github.com/kubernetes/minikube). Recommended for the
+     Linux hosts (also supports Mac/Windows).
+   - [Docker for Desktop](https://www.docker.com/products/docker-desktop).
+     Recommended for Mac/Windows.
 
 2. **Running on Google Kubernetes Engine (GKE)”** (~30 minutes) You will build,
    upload and deploy the container images to a Kubernetes cluster on Google
@@ -83,7 +88,7 @@ We offer three installation methods:
    option, you will use pre-built container images that are available publicly,
    instead of building them yourself, which takes a long time).
 
-### Option 1: Running locally with “Docker for Desktop”
+### Option 1: Running locally
 
 > 💡 Recommended if you're planning to develop the application or giving it a
 > try on your local cluster.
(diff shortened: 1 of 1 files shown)


ahmetb:
we strive to keep the "number of options" low to run this demo app.

so we should probably kill the "docker for desktop" option in favor of minikube. what do you think?

idvoretskyi:
> so we should probably kill the "docker for desktop" option in favor of minikube. what do you think?

Fine with that, but we'll need someone to test this with Minikube on Mac/Windows (I don't have them).

ahmetb:
I'm trying this, so far it crashed in the middle of `skaffold run` with no error once (4 CPUs, 3GiB RAM):

```
...
Step 3/16 : ENV PROJECT github.com/GoogleCloudPlatform/microservices-demo/src/productcatalogservice
 ---> Running in f1c075448b0f
 ---> 3055a5df5d85
Step 4/16 : WORKDIR /go/src/$PROJECT
 ---> Running in 48564497716d
 ---> 45c1487fb1f3
Step 5/16 : COPY Gopkg.* ./
 ---> a7fc148b57b6
Step 6/16 : RUN dep ensure --vendor-only -v
 ---> Running in 53d8a267940d
(1/21) Wrote github.com/googleapis/gax-go@v2.0.3
```

ahmetb:
Second build succeeded and it seems to be working OK (on 4cpus/3096MB).

Most notably, the last step of the local instructions needs to be changed from

> Visit http://localhost:80

to

> Run `minikube service frontend-external --url` and visit the URL

Sadly the URL looks much less intiutive, like `http://192.168.64.2:31594` (not only it's not on port `:80` as the Service spec defines, but also not exposed directly on localhost). Does Minikube offer anything to make this better?

But other than that, I can't seem to find any problems.

ahmetb:
Found another somewhat interesting problem.

Many users of this demo app are Googlers, and they all use the same browser with a corp proxy extension preinstalled.

When they go to the `http://192.168.64.2` like URL above, the proxy will treat the URL not local to the machine, but local to Google’s corporate network. (Docker for Mac doesn't have this problem as it uses `localhost` which never routes outside your machine.)

As a result, they will go to an error page that reads `Failed to connect to the backend.` because that IP is routed to our corp proxy. This will confuse Googlers quite a bit and we (particularly me) may get a lot of messages from new Googlers adopting this demo app for troubleshooting. 🤷‍♂️ I really wish there was a way to use `localhost` with Minikube.

idvoretskyi:
> I really wish there was a way to use localhost with Minikube.

@tstromberg do you have any suggestions on this?

README.md
```diff
@@ -68,11 +68,16 @@ Find **Protocol Buffers Descriptions** at the [`./pb` directory](./pb).
...
    and deploy microservices images to a single-node Kubernetes cluster running
-   on your development machine.
+   on your development machine. There are two options to run a Kubernetes
+   cluster locally for this demo:
+   - [Minikube](https://github.com/kubernetes/minikube). Recommended for the
+     Linux hosts (also supports Mac/Windows).
+   - [Docker for Desktop](https://www.docker.com/products/docker-desktop).
+     Recommended for Mac/Windows.
 
 2. **Running on Google Kubernetes Engine (GKE)”** (~30 minutes) You will build,
```
sdake: :wave: @idvoretskyi .  Note you didn't create this bug, however, for line 82 and others in a list, markdown just needs `1` and will autonumber ordered lists.

README.md
```diff
@@ -83,24 +88,39 @@ We offer three installation methods:
...
 
-1. Launch “Docker for Desktop”. Go to Preferences:
+1. Launch the local Kubernetes cluster with one of the following tools:
 
-   - choose “Enable Kubernetes”,
-   - set CPUs to at least 3, and Memory to at least 6.0 GiB
-   - on the "Disk" tab, set at least 32 GB disk space
+    - Launch Minikube (tested with Ubuntu Linux). Please, ensure that the
+       local Kubernetes cluster has at least:
+        - 4 CPU's
```
sdake: more accurate would be `4 Hyperthreads`

README.md
```diff
@@ -83,24 +88,39 @@ We offer three installation methods:
...
+        - 4GB RAM
```
sdake: to match the rest of the document `- 4.0 GiB memory`

README.md
```diff
@@ -83,24 +88,39 @@ We offer three installation methods:
...
+
+        To run a Kubernetes cluster with Minikube using the described configuration, please run:
+
+    ```shell
+    minikube start --cpus=4 --memory 4096
+    ```
+    
+    - Launch “Docker for Desktop” (tested with Mac/Windows). Go to Preferences:
+        - choose “Enable Kubernetes”,
+        - set CPUs to at least 3, and Memory to at least 6.0 GiB
```
sdake: I'd break this up into separate lines.

```
- Set CPUs to at minimum 3.
- Set memory to at minimum 6.0 GiB
- Set disk to 32 GiB using the `Disk` tab

README.md
idvoretskyi: @sdake oh, great to know :)

idvoretskyi:
Thank you @sdake, I'll address your comments.

README.md
ahmetb: I don't think minikube uses this terminology.

PR #210, opened 2019-05-02
 Add Stackdriver Profiler Java agent to adservice

kalyanac:
Changed the base image from openjdk:8-alpine to openjdk:8-slim. This causes increase in size of the final image from 180 MB to 350 MB.

Tested on GCP and non-GCP environments. On non-GCP environment, profiler will not initialize if proper credentials are not available.


src/adservice/Dockerfile
```diff
@@ -11,9 +11,14 @@ COPY . .
...
 RUN ./gradlew installDist
 
-FROM openjdk:8-alpine
-
-RUN apk add --no-cache libc6-compat
+FROM openjdk:8-slim
+
+# Download Stackdriver Profiler Java agent
+RUN apt-get update
+RUN apt-get install -qq wget
```
ahmetb: please do these

1. use -y to update
2. use -y to install as well
3. merge RUN steps so the intermediate resuls don't get cached
4. chain a step to remove the apt cache.

an example of this would look like:

```
RUN apt-get -y update && apt-get install -qqy \
    pkg-1 \
    pkg-2 \
    pkg-3
 && rm -rf /var/lib/apt/lists/*
```

src/adservice/Dockerfile
```diff
@@ -11,9 +11,14 @@ COPY . .
...
+RUN mkdir -p /opt/cprof && \
```
ahmetb: please remove the .tar.gz file after it's used in the same `RUN` because it doesn't need to be present in the image anymore after the pipe is complete.

```
RUN wget | tar && \
        rm -rf x.tar.gz
```

src/adservice/Dockerfile
kalyanac: done.

src/adservice/Dockerfile
kalyanac: done. 
With both suggested changes, image size came down to 330MB.

kalyanac:
continuous-integration/travis-ci/pr — The Travis CI build failed with:

```
remote: RESOURCE_EXHAUSTED Resource has been exhausted (e.g. check quota).
[type.googleapis.com/google.rpc.QuotaFailure]
violations {
  subject: "IP 104.154.113.151"
  description: "Short term bandwidth rate limit exceeded"
}
fatal: protocol error: bad pack header
: command failed: [git clone --recursive -v --progress https://go.googlesource.com/net /go/pkg/dep/sources/https---go.googlesource.com-net]: exit status 128
FATA[0132] build failed: building [shippingservice]: build artifact: running build: The command '/bin/sh -c dep ensure --vendor-only -v' returned a non-zero code: 1 
The command "skaffold build --profile travis-ci" exited with 1.
```

I dont see any way to restart it.

ahmetb:
Let's merge this since it’s enabled on a guarded code path. We’ll troubleshoot issues from this if it arises later, especially when deployed on Istio.

PR #218, opened 2019-05-21
Removed deprecated left-pad module dependency.

dineshnagarit:
In this PR I have removed the dependency of left-pad npm module, we no longer need to use due to following reasons.

1. https://www.npmjs.com/package/left-pad , This package has been deprecated and author suggested to use native String.prototype.padStart() javascript method.

2. It doesnt make sense to install a new module to perform just a single line of operation. In our case we were doing return `${m.units}.${m.nanos.toString().padStart(9,'0')} ${m.currency_code}`; instead of ${m.units}.${m.nanos.padStart(9,'0')} ${m.currency_code} .

### DIFF
--- a/src/currencyservice/client.js
+++ b/src/currencyservice/client.js
@@ -46,7 +46,7 @@ const request = {
 };
 
 function _moneyToString (m) {
-  return `${m.units}.${leftPad(m.nanos, 9, '0')} ${m.currency_code}`;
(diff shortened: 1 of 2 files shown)


dineshnagarit:
@ahmetb Kindly review the PR and let me know if there is any suggestions or queries. Thanks

Issue #222, opened 2019-06-05
loadgenerator crashing on start


iasinDev:
After running skaffold run, and getting pods you can see:

NAME                                     READY   STATUS             RESTARTS   AGE
adservice-57799fcdbd-twkk4               1/1     Running            0          78s
cartservice-8bd8479cf-p526k              1/1     Running            0          78s
checkoutservice-65c48b965f-gqcng         1/1     Running            0          78s
currencyservice-ff4b77bdc-vpqj2          1/1     Running            0          78s
emailservice-657fc6d6df-dldnb            1/1     Running            0          78s
frontend-6cb8846b67-vcgph                1/1     Running            0          77s
loadgenerator-79875bb976-tqn4d           0/1     CrashLoopBackOff   2          77s
paymentservice-54bc56fc78-8z2nf          1/1     Running            0          76s
productcatalogservice-7b4d46d7dd-7mf2x   1/1     Running            0          76s
recommendationservice-56db944d5c-2vck4   1/1     Running            0          74s
redis-cart-7894b7c796-gpmxf              1/1     Running            0          73s
shippingservice-7dc94f4cb5-w4qrg         1/1     Running            0          71s

Taking a look into loadgenerator:

$ kubectl logs loadgenerator-79875bb976-tqn4d
/bin/sh: 1: ./loadgen.sh: Permission denied

It's just make a fix in the Dockerfile, I'm going to submit a PR for this

iasinDev:
PR here: https://github.com/GoogleCloudPlatform/microservices-demo/pull/224

UsamaIslam:
open loadgen.sh in notepad++ (if using windows) and change the EOL to unix from Edit -> EOL Conversion -> UNIX (LF)

worked for me

PR #223, opened 2019-06-05
Fix https://github.com/GoogleCloudPlatform/microservices-demo/issues/222

iasinDev:
Fix https://github.com/GoogleCloudPlatform/microservices-demo/issues/222 by adding execution rights to the script


iasinDev:
I signed it!

PR #230, opened 2019-06-17
Frontend prometheus instrumentation

yuriatgoogle:
Added prometheus instrumentation to frontend code


yuriatgoogle:
pushed to wrong repo

Issue #199, opened 2019-03-26
Stackdriver exporters failing to detect GCE metadata api when run on Istio


ahmetb:
I think something's fundamentally wrong with Istio on GKE execution environment right now. This reproduces on HEAD with both Istio 1.0 and 1.1 (GKE Add-on).

Microservices are still failing to access metadata server after first 30s. They keep retrying. But when I exec into the pod, I can access the metadata API.

It's nondeterministic, sometimes registering tracing succeeds, but not profiling (basically similar metadata api calls trying to register if they're on GCE, if so trying to get a token).

This is a red flag if users are using Stackdriver Profiling/Debugger/Tracing or Opencensus exporers for Stackdriver on GKE w/ Istio.

Look at these logs:

`{"message":"failed to start profiler: failed to get the project ID from Compute Engine: Get http://169.254.169.254/computeMetadata/v1/project/project-id: dial tcp 169.254.169.254:80: connect: connection refused","severity":"warning","timestamp":"2019-03-08T22:54:23.649438404Z"}`

`{"message":"sleeping 10s to retry initializing stackdriver profiler","severity":"info","timestamp":"2019-03-08T22:54:23.649503248Z"}
{"message":"failed to initialize stackdriver exporter: stackdriver: no project found with application default credentials","severity":"info","timestamp":"2019-03-08T22:54:23.649674038Z"}`

`{"message":"sleeping 10s to retry initializing stackdriver exporter","severity":"info","timestamp":"2019-03-08T22:54:23.649687415Z"}`

`{"message":"failed to start profiler: failed to get the project ID from Compute Engine: Get http://169.254.169.254/computeMetadata/v1/project/project-id: dial tcp 169.254.169.254:80: connect: connection refused","severity":"warning","timestamp":"2019-03-08T22:54:33.649663529Z"}`

`{"message":"sleeping 20s to retry initializing stackdriver profiler","severity":"info","timestamp":"2019-03-08T22:54:33.650004528Z"}`

`{"message":"registered stackdriver tracing","severity":"info","timestamp":"2019-03-08T22:54:33.726637283Z"}` **TRACING REGISTRATION WORKED!**

`{"message":"failed to start profiler: failed to get the project ID from Compute Engine: Get http://169.254.169.254/computeMetadata/v1/project/project-id: dial tcp 169.254.169.254:80: connect: connection refused","severity":"warning","timestamp":"2019-03-08T22:54:53.650341347Z"}`

[...]

PR #234, opened 2019-07-09
Updates go services to google-cloud-go 0.40.0

askmeegs:
Fixes Stackdriver Profiler startup error (#199) for all Go services: `productcatalog`, `frontend`, `shippingservice`, and `checkoutservice`. 

3 changes for each of these: 

1. Upgrate google-cloud-go from `0.27.0` to `0.40.0`, which includes the fix for the Stackdriver profiler startup retry bug. 
2. Due to grpc incompatibility problems on that upgrade, updates grpc-go to `1.22.0`.  
3. Then due to [a known breaking change](https://github.com/grpc/grpc-go/issues/2313) in the Health interface for grpc go post 1.14.0, adds a no-op `Watch()` function for each of the services.   

Confirmed that now, Profiler can correctly start up within 2 retries, with Istio enabled, on each of the 4 services. Example, shippingservice:

```
shippingservice-949db8b5c-mr5xb server {"message":"registered Stackdriver tracing","severity":"info","timestamp":"2019-07-09T15:38:45.415591318Z"}
shippingservice-949db8b5c-mr5xb server {"message":"Registered default server views","severity":"info","timestamp":"2019-07-09T15:38:45.415686246Z"}
shippingservice-949db8b5c-mr5xb server {"message":"started Stackdriver profiler","severity":"info","timestamp":"2019-07-09T15:38:45.418201652Z"}
shippingservice-949db8b5c-mr5xb server {"message":"[GetQuote] received request","severity":"info","timestamp":"2019-07-09T15:38:45.634426144Z"}
```

### DIFF
--- a/src/productcatalogservice/server.go
+++ b/src/productcatalogservice/server.go
@@ -240,6 +240,10 @@ func (p *productCatalog) Check(ctx context.Context, req *healthpb.HealthCheckReq
 	return &healthpb.HealthCheckResponse{Status: healthpb.HealthCheckResponse_SERVING}, nil
 }
 
+func (p *productCatalog) Watch(req *healthpb.HealthCheckRequest, ws healthpb.Health_WatchServer) error {
+	return nil
+}
+
 func (p *productCatalog) ListProducts(context.Context, *pb.Empty) (*pb.ListProductsResponse, error) {
 	time.Sleep(extraLatency)
 	return &pb.ListProductsResponse{Products: parseCatalog()}, nil
--- a/src/productcatalogservice/server.go
+++ b/src/productcatalogservice/server.go
@@ -241,7 +241,7 @@ func (p *productCatalog) Check(ctx context.Context, req *healthpb.HealthCheckReq
 }
 
 func (p *productCatalog) Watch(req *healthpb.HealthCheckRequest, ws healthpb.Health_WatchServer) error {
-	return nil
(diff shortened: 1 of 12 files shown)


src/shippingservice/main.go
```diff
@@ -90,6 +90,10 @@ func (s *server) Check(ctx context.Context, req *healthpb.HealthCheckRequest) (*
 	return &healthpb.HealthCheckResponse{Status: healthpb.HealthCheckResponse_SERVING}, nil
 }
 
+func (s *server) Watch(req *healthpb.HealthCheckRequest, ws healthpb.Health_WatchServer) error {
+	return nil
```
ahmetb: Can we return a proper grpc error here instead so that if somehow this is invoked in the future, we don't ignore it. http://avi.im/grpc-errors/#go

    return status.Errorf(codes.Unimplemented, "health check via Watch not implemented")

src/productcatalogservice/server.go
```diff
@@ -240,6 +240,10 @@ func (p *productCatalog) Check(ctx context.Context, req *healthpb.HealthCheckReq
 	return &healthpb.HealthCheckResponse{Status: healthpb.HealthCheckResponse_SERVING}, nil
 }
 
+func (p *productCatalog) Watch(req *healthpb.HealthCheckRequest, ws healthpb.Health_WatchServer) error {
+	return nil
```
ahmetb: ditto as below

PR #238, opened 2019-07-22
Mc errors

mikecohen:
supports generating 500's via HTTP_500_PERCENT environment variable and modifies locust to not use active connections.


mikecohen:
whoops, didn't intend to do the PR against this...sorry

PR #239, opened 2019-07-31
Set environment variable to enable Profiler

mtwo:


mtwo:
@kalyanac FYI

ahmetb:
I forgot why we didn’t enable it by default. @kalyanac perhaps remembers. 
Is it enabled by default in other services?

mtwo:
Yeah on the non-Python ones I think

________________________________

mtwo:
@kalyanac do the Profiler agents in other languages also check for the environment variable?

mtwo:
Confirmed from using hipstershop that it's enabled by default in other languages

kalyanac:
> I forgot why we didn’t enable it by default. @kalyanac perhaps remembers.

@ahmetb, you wanted the code set to disabled by default and let users enable manually. 

> Is it enabled by default in other services?
It may be enabled by default on some. Java is on by default since there is no way to control it from within the code.

mtwo:
Thanks Kalyana, I'd forgotten about that. Right now Profiler appears to be enabled by default on some services and not others. I'd suggest that we enable it by default on all, as I've had to re-enable it on countless GCP demos that teams have set up where they didn't know that it had to be enabled.

kalyanac:
I also would prefer to have Profiler enabled by default. @ahmetb had some valid objections the last time we discussed this. I think the retry and abort logic should address most of those.

ahmetb:
I forgot the details of our conversation back in the day.
I think it was an issue limited to Istio, right?
Also did you end up enabling the retry/give-up logic?

kalyanac:
Yes, retry/give-up is implemented.

DanSanche:
Can you make the same change to the recommendation service in this PR?

(I'd also vote that it should be enabled by default to be consistent with the non-Python services)

PR #243, opened 2019-08-23
Update handlers.go

marcfong:
Changed httpError for failing to complete order to "http.StatusBadRequest"


ahmetb:
I don't think this is necessarily accurate. If the rpc fails, it's not badrequest.

We might have issues throughout services like this that we don't distinguish between input problems vs service problems. What's your driver for changing this?

Issue #10, opened 2018-08-03
cartservice: move tests/ dir out of root


ahmetb:
`/tests/cartservice` should go be moved to `/src/cartservice/tests/cartservice`.

djmailhot:
> Does vs or vscode’s .NET/C# support understand this approach and still able to run the tests from the IDE?
- @ahmetb

DanSanche:
@djmailhot Do you have any objections to this? I'd say IDE tests aren't too important

djmailhot:
@DanSanche Apologies, I'm confused.  What is your question again?

DanSanche:
I think ahmet was suggesting moving the test from `/tests/cartservice` to `/src/cartservice/tests/cartservice` to be consistent with the other services. Any objections? Were you saying that would break IDE tests or something?

djmailhot:
> `/tests/cartservice` should go be moved to `/src/cartservice/tests/cartservice`.

Turns out that you [can't nest one CS project within another](https://github.com/xunit/xunit/issues/1467#issuecomment-333404867).

djmailhot:
My PR is useless, so I'll delete it and release this issue

PR #248, opened 2019-09-25
Move /tests/cartservice to /src/cartservice/tests/cartservice

djmailhot:
[cartservice: move tests/ dir out of root](https://github.com/GoogleCloudPlatform/microservices-demo/issues/10)


ahmetb:
Does vs or vscode’s .NET/C# support understand this approach and still able to run the tests from the IDE?

djmailhot:
Turns out that you [can't nest one CS project within another](https://github.com/xunit/xunit/issues/1467#issuecomment-333404867).

Closing this PR as infeasible.

PR #250, opened 2019-09-26
Update application architecture diagram with Languages and Protocols

rafik8:
Update application architecture diagram to include Languages and Protocols.
Source diagram is under docs/diagrams.


ahmetb:
Thanks but we likely can't take this because we use Lucidchart to maintain the diagram. It would be better if you make a proposal first next time before going the trouble of completing the work.

rafik8:
Hi Ahmed, No worries. I am using the application for a workshop that I am already working on so the job was already done.

PR #254, opened 2019-10-16
use python 3.7, as python:3-slim is now using python 3.8.

jsirianni:
The email service fails to build because it requires python 3.7, not 3.8. python:3-slim now ships with 3.8.

### DIFF
--- a/src/emailservice/Dockerfile
+++ b/src/emailservice/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3-slim as base
+FROM python:3.7-slim as base
 
 FROM base as builder
 


jsirianni:
@googlebot I signed it!

ahmetb:
An obvious question is why is this change required? Anything wrong with 3.8?

jsirianni:
This line in the Dockerfile fails:
```
COPY --from=builder /usr/local/lib/python3.7/ /usr/local/lib/python3.7/
```

ahmetb:
Looks like that optimization is finally hurting us.

We have python:2.7-slim in another image (https://github.com/GoogleCloudPlatform/microservices-demo/blob/2177813c44a83f3f6bc77e14d88accf986dd3a4c/src/recommendationservice/Dockerfile#L1), so I think this is change is ok to make.

Thanks.

Issue #269, opened 2019-12-09
Load generator fails with Istio strict mTLS


howardjohn:
The init container is trying to call the frontend before Envoy has started. This will send plaintext traffic, which will fail since the frontend is requiring mTLS traffic only.

ahmetb:
🤔 Hmm we can remove the init container maybe (IIRC it only makes sure frontend comes up), instead maybe we can just fail the requests. That'll show some failures in load generator output, but it'll eventually still send load.

howardjohn:
@ahmetb I would suggest either that or have load gen crashloop until frontend comes up (not sure how feasible that is). That is how `fortio` works as far as I know.

ahmetb:
yeah, we can do something like a liveness probe on the loadgen to curl the frontend.

howardjohn:
I was thinking as part of the code for loadgen. How fortio works (referring to this since I think their solution makes sense here) is that if you do 

`load foo.com`

It will send a single request to `foo.com`. If it fails, it will exit with some exit code. If it passes, it will then send a bunch of load to foo.com and record success/failures.

Either way works though

PR #270, opened 2019-12-10
Remove init container on load gen

howardjohn:
Partial fix for https://github.com/GoogleCloudPlatform/microservices-demo/issues/269

Perfectly understand if we want a fix that preserves the functionality. I am giving a demo with strict mtls so will be using this PR, so figured I would submit it at least

### DIFF
--- a/kubernetes-manifests/loadgenerator.yaml
+++ b/kubernetes-manifests/loadgenerator.yaml
@@ -28,17 +28,6 @@ spec:
     spec:
       terminationGracePeriodSeconds: 5
       restartPolicy: Always
-      initContainers:
-      - name: wait-frontend
-        image: alpine:3.6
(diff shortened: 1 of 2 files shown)


ahmetb:
Let's revert manifest/**, that's meant to be updated when we have a new tag.

ahmetb:
Let's revert release/**, that's meant to be updated when we have a new tag.

askmeegs:
Suggest maybe closing this one in favor of https://github.com/GoogleCloudPlatform/microservices-demo/pull/279, which removes the init container but also adds a single curl to the beginning of the loadgen src? 
(sorry, just saw this @howardjohn !)

askmeegs:
Closing in favor of #279 which was merged.

PR #279, opened 2019-12-15
Standardizes loadgen behavior for Istio mTLS=STRICT

### LINKED ISSUE #211: Istio mTLS now supports readiness/liveness probes - update README
Current [README instructs](https://github.com/GoogleCloudPlatform/microservices-demo#optional-deploying-on-a-istio-installed-gke-cluster) to remove the liveness/readiness probes from deployments, if the user has installed Istio with STRICT mTLS. 

OSS Istio 1.1 [now supports](https://istio.io/about/notes/1.1/#security) probes in mTLS, so we may want to:

1. test that this is the case 
2. modify the README accordingly

askmeegs:
Closes #211, #269 

### Changelog 

Updates the loadgen and frontend when Istio mTLS is set to `STRICT`.

1. Fixes the case where the loadgen tries to call the frontend before the client-side Envoy starts, causing non mTLS traffic --> curl fails --> loadgen never starts. Now, loadgen init container logic (curl frontend) moved into the first line of the `main` script, before locust starts. This way, the loadgen will `CrashLoop` if it can't reach the frontend, and will retry instead of getting stuck. 
1. Added pod annotations for frontend, loadgen for [`sidecar.istio.io/rewriteAppHTTPProbers: "true"`](https://istio.io/docs/ops/configuration/mesh/app-health-check/#use-annotations-on-pod) -- this will rewrite HTTP probes to be TLS, to work when mTLS is set to STRICT.    
1. Updates README instructions to remove liveness/readiness probes for mTLS -- these manifests should now work with any Istio mTLS settings. 

### How to test 

1. Create a regular GKE cluster (I used 4 nodes, n1-standard-4)
2. Install OSS Istio. I used [this script](https://gist.github.com/askmeegs/bdc283dc8e39b167795aac4b90ea5918), and set `ISTIO_VERSION` to 1.4.2 
3. Enable mesh-wide STRICT mtls -- apply [this yaml](https://gist.github.com/askmeegs/30ee3256de606646388b80184e2bb1c5)
4. update `kubernetes-manifests/` loadgen image to`gcr.io/megandemo/loadgen:failcurl` 
5. deploy hipstershop on the cluster. everything should start normally, liveness/readiness probes should work, locust starts 
6. Trigger the failing state: `kubectl delete deployment frontend` 
7. delete the loadgen pod to re-kick 
8. `kubectl logs <loadgen-pod> -c main`. You should see: `Error: Loadgen cannot reach frontend, exiting`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,3 @@ pkg/
 .idea
 .skaffold-*.yaml
 .kubernetes-manifests-*/
-istio-1.*/
-install-istio.sh
-mtls.yaml
-istio.yaml
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,7 @@ pkg/
 .idea
 .skaffold-*.yaml
 .kubernetes-manifests-*/
+istio-1.*/
+install-istio.sh
+mtls.yaml
+istio.yaml
--- a/release/kubernetes-manifests.yaml
+++ b/release/kubernetes-manifests.yaml
@@ -451,22 +451,9 @@ spec:
     spec:
       terminationGracePeriodSeconds: 5
       restartPolicy: Always
-      # initContainers:
-      # - name: wait-frontend
-      #   image: alpine:3.6
-      #   command: ['sh', '-c', 'set -x;  apk add --no-cache curl &&
-      #     until timeout -t 2 curl -f "http://${FRONTEND_ADDR}"; do
-      #       echo "waiting for http://${FRONTEND_ADDR}";
-      #       sleep 2;
-      #     done;']
-      #   env:
-      #   - name: FRONTEND_ADDR
-      #     value: "frontend:80"
       containers:
       - name: main
-#        image: gcr.io/google-samples/microservices-demo/loadgenerator:v0.1.2
         image: gcr.io/megandemo/loadgen:failcurl
-        imagePullPolicy: Always
         env:
         - name: FRONTEND_ADDR
           value: "frontend:80"
--- a/release/kubernetes-manifests.yaml
+++ b/release/kubernetes-manifests.yaml
@@ -453,7 +453,7 @@ spec:
       restartPolicy: Always
       containers:
       - name: main
-        image: gcr.io/megandemo/loadgen:failcurl
+        image: gcr.io/google-samples/microservices-demo/loadgenerator:v0.1.2
         env:
         - name: FRONTEND_ADDR
           value: "frontend:80"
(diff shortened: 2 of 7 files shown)


src/loadgenerator/loadgen.sh
```diff
@@ -18,10 +18,19 @@
...
     echo >&2 "FRONTEND_ADDR not specified"
     exit 1
 fi
 
 set -x
+
+# if one request to the frontend fails, then exit
+STATUSCODE=$(curl --silent --output /dev/stderr --write-out "%{http_code}" http://${FRONTEND_ADDR})
+if test $STATUSCODE -ne 200; then
+    echo "Error: cannot reach frontend, exiting"
```
ahmetb: might be wise to add $STATUSCODE to error msg

src/loadgenerator/loadgen.sh
```diff
@@ -18,10 +18,19 @@
 set -e
 trap "exit" TERM
 
-if [[ -z "${FRONTEND_ADDR}" ]]; then
+if [ -z "${FRONTEND_ADDR}" ]; then
```
ahmetb: there should be really no difference between [ and [[ in this case.

release/kubernetes-manifests.yaml
```diff
@@ -450,9 +450,9 @@ spec:
       initContainers:
       - name: wait-frontend
         image: alpine:3.6
-        command: ['sh', '-c', 'set -x;  apk add --no-cache curl && 
-          until timeout -t 2 curl -f "http://${FRONTEND_ADDR}"; do 
-            echo "waiting for http://${FRONTEND_ADDR}"; 
+        command: ['sh', '-c', 'set -x;  apk add --no-cache curl &&
+          until timeout -t 2 curl -f "http://${FRONTEND_ADDR}"; do
+            echo "waiting for http://${FRONTEND_ADDR}";
```
ahmetb: do we still need to update this?
(I assume you didn't mean to touch this file, as it is autogenerated.)

.gitignore
```diff
@@ -9,3 +9,7 @@ pkg/
 .idea
 .skaffold-*.yaml
 .kubernetes-manifests-*/
+istio-1.*/
```
ahmetb: did you mean to touch this file?

release/kubernetes-manifests.yaml
askmeegs: I had accidentally updated the release/ manifests, then reverted that, I think that's just my linter.

.gitignore
askmeegs: nope, fixed

Issue #283, opened 2020-01-16
cartservice and loadgenerator crash at start


Nusserdt:
I follow the installation steps from option 3. I had already a running kubernetes cluster. So I only execute:

`kubectl apply -f ./release/kubernetes-manifests.yaml`

When I evaluate the result with `kubectl get pods` I get:

```
NAME                                     READY   STATUS             RESTARTS   AGE
adservice-55f9757757-9tb2h               1/1     Running            0          16m
cartservice-684bb46b44-b6dvk             0/1     CrashLoopBackOff   8          16m
checkoutservice-6fcc84467f-gm79f         1/1     Running            0          16m
currencyservice-6c7c479d45-qfnl6         1/1     Running            0          16m
emailservice-8dd9b76cc-7j6zj             1/1     Running            0          16m
frontend-7d8cfc75b5-dqnkc                1/1     Running            0          16m
loadgenerator-5db67d555-fq42k            0/1     CrashLoopBackOff   7          16m
paymentservice-84ffc75c55-mzwfx          1/1     Running            0          16m
productcatalogservice-d564bdf4c-bch2r    1/1     Running            0          16m
recommendationservice-76598d5889-p9qhm   1/1     Running            0          16m
redis-cart-5f59546cdd-rqqdf              1/1     Running            0          16m
shippingservice-b6db65f7f-t54ng          1/1     Running            0          16m
```
`cartservice` and `loadgenerator` are not able to start.

---

### Logs

`kubectl logs cartservice-684bb46b44-b6dvk`

```
Started as process with id 1
Reading host address from LISTEN_ADDR environment variable
Reading cart service port from PORT environment variable
Reading redis cache address from environment variable REDIS_ADDR
Connecting to Redis: redis-cart:6379,ssl=false,allowAdmin=true,connectRetry=5
StackExchange.Redis.RedisConnectionException: It was not possible to connect to the redis server(s). UnableToConnect on redis-cart:6379/Interactive, Initializing/NotStarted, last: NONE, origin: BeginConnectAsync, outstanding: 0, last-read: 5s ago, last-write: 5s ago, keep-alive: 180s, state: Connecting, mgr: 10 of 10 available, last-heartbeat: never, global: 10s ago, v: 2.0.601.3402
[...]

PR #286, opened 2020-01-22
Improve startup using InitContainers

DanSanche:
Some users were confused by the crashes in cartservice and loadgenerator as they wait for the redis-cart service to be ready (https://github.com/GoogleCloudPlatform/microservices-demo/issues/283). This PR fixes that using InitContainers to block pod creation until dependencies are ready

I also improved the readiness probe of the redis-cart, so it only shows as ready when it can respond to a redis-cli ping

### DIFF
--- a/kubernetes-manifests/cartservice.yaml
+++ b/kubernetes-manifests/cartservice.yaml
@@ -26,6 +26,13 @@ spec:
         app: cartservice
     spec:
       terminationGracePeriodSeconds: 5
+      initContainers:
+      - name: init-db-ready
+        image:  redis:alpine
+        command: ['/bin/sh', '-c']
+        args:
+          - until redis-cli -h redis-cart -p 6379 ping; do echo waiting for redis-cart; sleep 2; done
+          #- for i in {1..100}; do sleep 1; if redis-cli -h redis-cart -p 6379 ping; then exit 0; fi; done; exit 1
       containers:
       - name: server
         image: cartservice
--- a/kubernetes-manifests/cartservice.yaml
+++ b/kubernetes-manifests/cartservice.yaml
@@ -32,7 +32,6 @@ spec:
         command: ['/bin/sh', '-c']
         args:
           - until redis-cli -h redis-cart -p 6379 ping; do echo waiting for redis-cart; sleep 2; done
-          #- for i in {1..100}; do sleep 1; if redis-cli -h redis-cart -p 6379 ping; then exit 0; fi; done; exit 1
       containers:
       - name: server
         image: cartservice
--- a/kubernetes-manifests/loadgenerator.yaml
+++ b/kubernetes-manifests/loadgenerator.yaml
@@ -32,7 +32,9 @@ spec:
       initContainers:
       - name: init-frontend-ready
         image: busybox:1.28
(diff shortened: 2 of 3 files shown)


kubernetes-manifests/cartservice.yaml
```diff
@@ -26,6 +26,12 @@ spec:
         app: cartservice
     spec:
       terminationGracePeriodSeconds: 5
+      initContainers:
```
ahmetb: I'm -1 on adding an initContainer. While it fixes the immediate problem, it adds hard dependency to redis to start up. This is not regarded as a good practice in Kubernetes overall. It's a hack.

Instead we can fix cartservice's health method to not to talk to redis.

I think we should solve the root problem here. cartservice is almost always gonna start slower than a redis container (it's larger in size, and uses .net which has a much slower startup time than a C program).

kubernetes-manifests/loadgenerator.yaml
```diff
@@ -29,6 +29,12 @@ spec:
     spec:
       terminationGracePeriodSeconds: 5
       restartPolicy: Always
+      initContainers:
+      - name: init-frontend-ready
+        image: busybox:1.28
+        command: ['/bin/sh', '-c']
+        args:
+          - until  [[ "$(wget --spider -S http://frontend 2>&1 | grep HTTP/)" == "  HTTP/1.1 200 OK" ]]; do echo waiting for frontend; sleep 2; done
```
ahmetb: We actually had this. But we had to remove it because of istio.

ahmetb: See #279.
https://github.com/GoogleCloudPlatform/microservices-demo/commit/8cfb88bf2f4eb1843b08b45138b4020b5681493f#diff-cbd8d0a4b56cb8944c9951eac8e9104f

Let's not add it again.

kubernetes-manifests/redis.yaml
```diff
@@ -32,8 +32,8 @@ spec:
         - containerPort: 6379
         readinessProbe:
           periodSeconds: 5
-          tcpSocket:
-            port: 6379
+          exec:
+            command: ['/bin/sh', '-c', 'redis-cli -h $(hostname) ping']
```
ahmetb: I think you can probably use localhost as hostname, in that case you would not need /bin/sh -c

also, why make this change? does redis somehow start up on port 6379, and not be ready sometimes?

kubernetes-manifests/loadgenerator.yaml
DanSanche: @askmeegs Users are interpreting the load generator crash loop as a bug. Is there anything we could do to make the InitContainer play nice with mTLS? Or is it not worth the trouble?

kubernetes-manifests/cartservice.yaml
DanSanche: @simonz130 thoughts? Can we just remove the health check? (https://github.com/GoogleCloudPlatform/microservices-demo/blob/master/src/cartservice/HealthImpl.cs)

Removing it seems to work fine for me, but I'm not too familiar with C#

kubernetes-manifests/redis.yaml
DanSanche: I had a theory that the redis was reporting ready before it was fully ready (based on the cartservice crashloops), but I'm not sure if that's true anymore.

I thought I'd keep this since it feels more accurate to ping with redis-cli directly, I'm not sure if there are any actual functional differences, so I could take or leave this change

PR #288, opened 2020-01-27
GitHub Actions smoke tests

DanSanche:
Replaces Travis with GitHub Actions smoke tests. Tests run on each commit to master and each pull request, and do the following:
- build containers
- deploy containers to kind cluster
- deploy manifests in releases/ directory
- ensure all pods reach ready state
- ensure frontend responds HTTP 200

I also added a README with instructions for setting up a worker VM to run actions


.github/workflows/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Setup
+- workloads run using [GitHub self-hosted runners](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-self-hosted-runners)
+- project admins maintain a private Google Compute Engine VM for running tests
+  - VM should be at least n1-standard-4 with 50GB persistent disk
+  - instructions for setting up the VM can be found in repo settings under "Actions"
+  - to set up dependencies, run the following commands:
+    ```
+    # install kubectl
+    sudo apt-get install kind kubectl
```
ahmetb: which one installs kind? this or below?

.github/workflows/README.md
```diff
@@ -0,0 +1,55 @@
...
+    # install kind
+    curl -Lo ./kind "https://github.com/kubernetes-sigs/kind/releases/download/v0.7.0/kind-$(uname)-amd64"
+    chmod +x ./kind
+    sudo mv ./kind /usr/local/bin
+    # install skaffold
+    curl -Lo skaffold https://storage.googleapis.com/skaffold/releases/latest/skaffold-linux-amd64
```
ahmetb: you can perhaps do `&& \` for the related commands and separate them into code blocks for better readability.

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+        RESULT=" "
+        while [[ "$RESULT" != "  HTTP/1.1 200 OK" ]]; do
+          sleep 1
+          RESULT=$(kubectl exec deployments/frontend -- sh -c "wget --spider -S "http://frontend" 2>&1 | grep 'HTTP/'")
+          echo "front end response: $RESULT"
+        done
+        if [[ "$RESULT" != "  HTTP/1.1 200 OK" ]]; then 
+          exit 1
+        fi
+  test-release:
```
ahmetb: tbh I am not sure about the value of testing `release/` on every commit

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+    - uses: actions/checkout@v2
+    - name: Clean Cluster
+      run: |
+        kubectl delete all --all --wait=true --force --grace-period=0
+        sleep 5
+    - name: Deploy From Source
+      run: |
+        skaffold run
+    - name: Wait For Pods
+      run: |
```
ahmetb: you might wanna add a timeout to this step.

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+        sleep 60
+        kubectl get pods
+        echo waiting for ready state...
+        kubectl wait --for=condition=ready pod -l app=adservice --timeout=500s
```
ahmetb: If kubectl wait has an option to wait on `deployment` rather than `pod` (I think it does), that would be preferable.

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+        kubectl wait --for=condition=ready pod -l app=cartservice --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=checkoutservice --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=currencyservice --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=emailservice --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=frontend --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=loadgenerator --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=paymentservice --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=productcatalogservice --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=recommendationservice --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=shippingservice --timeout=500s
```
ahmetb: I'm 99% sure this command will not fail the step (and therefore the CI job) because you don't have && between these commands or have `set -e` set on this shell.

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+    steps:
+    - name: Check Test Cluster
+      run: |
+        if kubectl get nodes; then
+          kubectl get nodes
+        else
+          echo "test cluster not active. Attempting to create through kind..."
+          kind delete cluster 2>/dev/null
+          kind create cluster
+          kubectl get nodes
```
ahmetb: need && between these, otherwise I'm fairly certain this step won't fail if a middle command fails.

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+        kubectl wait --for=condition=ready pod -l app=productcatalogservice --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=recommendationservice --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=shippingservice --timeout=500s
+    - name: Test HTTP
+      timeout-minutes: 5
+      run: |
+        RESULT=" "
+        while [[ "$RESULT" != "  HTTP/1.1 200 OK" ]]; do
+          sleep 1
+          RESULT=$(kubectl exec deployments/frontend -- sh -c "wget --spider -S "http://frontend" 2>&1 | grep 'HTTP/'")
```
ahmetb: since we're on localhost, don't we get an IP address to the frontend somehow with kind?

.github/workflows/smoke-tests.yml
ahmetb: also if you eliminate this, you will be avoiding duplication.

.github/workflows/README.md
DanSanche: good catch, it's the second one. I'll remove this

.github/workflows/smoke-tests.yml
DanSanche: We could break it up into a separate workflow to run only on /release branches if that's preferable. The added cost of running it each time isn't too high either though, and it could be useful for a periodic sanity test (presumably we won't be committing to master too often)

.github/workflows/smoke-tests.yml
DanSanche: GitHub Actions runs with `set -e` by default.

.github/workflows/smoke-tests.yml
DanSanche: GitHub Actions will fail if any command fails (just re-tested to be sure)

.github/workflows/smoke-tests.yml
ahmetb: you got me there. when I used earlier it wasn't the case.
https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions
shows `-eo pipefail` that's good.

.github/workflows/smoke-tests.yml
ahmetb: The reason I say that is because no pods will be created if there's something more fundamental failing. So that might succeed.

However, Deployment will be there as it's part of `apply`, so it's better to check its readiness.

.github/workflows/smoke-tests.yml
ahmetb: We've hit this duplication on Krew and it was nasty. We ended up merging the release.yml and ci.yml (same job runs both on PRs and tags).

I'd say "periodic sanity test" isn't a great argument mostly because if/when we change something, we're now potentially testing a "broken" old release/**.yml file.

And we're supposed to update release/**.yml file only when tagging releases. So it would be good to leave release/** out of CI.

.github/workflows/smoke-tests.yml
DanSanche: I believe kind just shows "pending" for the external-ip field (https://github.com/kubernetes-sigs/kind/issues/702)

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+        kubectl wait --for=condition=ready pod -l app=checkoutservice --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=currencyservice --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=emailservice --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=frontend --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=loadgenerator --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=paymentservice --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=productcatalogservice --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=recommendationservice --timeout=500s
+        kubectl wait --for=condition=ready pod -l app=shippingservice --timeout=500s
+    - name: Test HTTP
```
ahmetb: you can call this "Smoke test" and change the actual job name to "Continuous integration"

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+        kubectl delete all --all --wait=true --force --grace-period=0
+        sleep 5
+    - name: Deploy Release Manifests
+      run: |
+        kubectl apply -f ./release/kubernetes-manifests.yaml
+    - name: Wait For Pods
+      run: |
+        sleep 60
+        kubectl get pods
+        echo waiting for ready state...
```
ahmetb: why need sleep 60 if the stuff below is already is gonna wait?
also probably don't need `kubectl get pods`.
also probably don't need echo, as you can do `set -x` to see what's being run.

.github/workflows/README.md
DanSanche: done

.github/workflows/smoke-tests.yml
DanSanche: That seems to work fine too. Done

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+        fi
+  test-source:
+    runs-on: self-hosted
+    needs: setup
+    steps:
+    - uses: actions/checkout@v2
+    - name: Clean Cluster
+      run: |
+        kubectl delete all --all --wait=true --force --grace-period=0
+        sleep 5
```
ahmetb: This is not good. `all` indeed will not clean everything.
If the costs aren't too high, I recommend deleting a cluster and creating again.
This almost certainly is going to leak something.
It's also a security problem to reuse a cluster between patches.

.github/workflows/README.md
```diff
@@ -0,0 +1,59 @@
+# GitHub Actions Workflows
+
+## Setup
+- workloads run using [GitHub self-hosted runners](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-self-hosted-runners)
+- project admins maintain a private Google Compute Engine VM for running tests
+  - VM should be at least n1-standard-4 with 50GB persistent disk
+  - instructions for setting up the VM can be found in repo settings under "Actions"
```
ahmetb: it's worth noting that the VM should be run with an empty (permissionless) service account, otherwise anyone who can send a PR can steal an `Editor` token for project.

.github/workflows/smoke-tests.yml
ahmetb: yeah, but I suspect there's a nodePort for each Service that you can directly curl.
I bet kind has a way to query a Service from the linux machine where kind is installed? :)

.github/workflows/smoke-tests.yml
DanSanche: hmm that's a good point. Yeah I'll remove the test-release then to make the tests simpler

.github/workflows/smoke-tests.yml
DanSanche: The `sleep/kubectl get pods` was mostly just to print the pod status to the console so you can see if anything looks out of place

This is less necessary if we use `set -x` though, so I may just switch to that

.github/workflows/smoke-tests.yml
DanSanche: creating a new cluster isn't too difficult if it's better practice. Done

.github/workflows/README.md
DanSanche: good point. added

.github/workflows/smoke-tests.yml
DanSanche: The research I've done is saying this is [more complicated than it looks](https://github.com/kubernetes-sigs/kind/issues/99#issuecomment-464614894), since we have to route through the extra layer of docker networking. If you're able to point me in the direction of a cleaner solution than this I'm happy to look into it though

ahmetb: That's fine then no worries.

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,51 @@
...
+    branches:
+      - master
+jobs:
+  run-tests:
+    runs-on: self-hosted
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Cluster
+      run: |
+        kind delete cluster 2>/dev/null
```
ahmetb: you can do `|| true` if you don't want it to fail. also consider adding set +x; so that you can tell what's happening

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,51 @@
+name: "Continuous Integration"
```
ahmetb: consider renaming to ci.yml

ahmetb:
Looks pretty good. Feel free to merge.

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,51 @@
...
+        kubectl wait --for=condition=available --timeout=500s deployment/productcatalogservice
+        kubectl wait --for=condition=available --timeout=500s deployment/recommendationservice
+        kubectl wait --for=condition=available --timeout=500s deployment/shippingservice
+    - name: Smoke Test
+      timeout-minutes: 5
+      run: |
+        RESULT=" "
+        while [[ "$RESULT" != "  HTTP/1.1 200 OK" ]]; do
+          sleep 1
+          RESULT=$(kubectl exec deployments/frontend -- sh -c "wget --spider -S "http://frontend" 2>&1 | grep 'HTTP/'")
```
ahmetb: unless I'm mistaken, wget already returns non-zero exit code (which propagates to `sh`, which propagates to `kubectl exec`). doesn't something like `wget -qO/dev/null [...]` work already?

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,51 @@
...
+        kubectl wait --for=condition=available --timeout=500s deployment/cartservice
+        kubectl wait --for=condition=available --timeout=500s deployment/checkoutservice
+        kubectl wait --for=condition=available --timeout=500s deployment/currencyservice
+        kubectl wait --for=condition=available --timeout=500s deployment/emailservice
+        kubectl wait --for=condition=available --timeout=500s deployment/frontend
+        kubectl wait --for=condition=available --timeout=500s deployment/loadgenerator
+        kubectl wait --for=condition=available --timeout=500s deployment/paymentservice
+        kubectl wait --for=condition=available --timeout=500s deployment/productcatalogservice
+        kubectl wait --for=condition=available --timeout=500s deployment/recommendationservice
+        kubectl wait --for=condition=available --timeout=500s deployment/shippingservice
```
ahmetb: now here's what I'm thinking, if we add a deployment, we have to add it here, too.

instead can we do:


```
for d in $(kubectl get deployments -o=name); do
  kubectl wait "$d"
done
```

since we expect all deployments to succeed?

.github/workflows/smoke-tests.yml
aojea: if is a Linux machine, assuming you have access to the docker network and let's say one of the docker nodes has the IP 172.17.0.3, you can access services in a different way:
* a nodeport service, using one of the docker nodes ip address (i.e. 172.17.0.3:31788)
* a clusterIP service, adding a route to the service IP network or cluster IP address through one of the docker node IPs (i.e. ip route add 10.96.0.10 via 172.17.0.3)
* playing with the external-ip and LB as explained in the kind issue that you've mentioned

DanSanche: I'd rather not add any new services or routes just for this test. I'd consider changing this test if there's a better one line option, but this so far seems like the least flaky option to me

.github/workflows/smoke-tests.yml
DanSanche: That is an option as well. I thought it might be better to be explicit which services we're expecting in case one was accidentally removed (until we have more thorough tests in place). I'm fine with either method though

DanSanche:
@ahmetb Sorry about the delay, I'm ready to finalize this up now. It looks like I don't have the permissions to enable Actions on this repo though (I was testing on a fork). Are you able to give me access so I can enable the tests?

ahmetb:
You should have now.

.github/workflows/smoke-tests.yml
DanSanche: Hmm that's a good point. That way is cleaner, but we'd lose the retry loop. I might do some testing on my fork to see if it's flaky

PR #289, opened 2020-01-28
Fix Local CartStore emptycart()

askmeegs:
When using a local cart store, the emptyCart function does not seem to work --  and  the loadgen keeps adding items, causing the cart total to exponentially increase. 
Screen Shot 2020-01-28 at 8 51 52 AM

I'm not 100% sure, but I think the `emptyCart` variable being passed around / reassigned, on emptyCart() was the issue. When I changed the emptyCart() function to instead initialize a clean `Hipstershop.Cart`, the problem went away.

### DIFF
--- a/src/cartservice/cartstore/LocalCartStore.cs
+++ b/src/cartservice/cartstore/LocalCartStore.cs
@@ -65,7 +65,7 @@ namespace cartservice.cartstore
         public Task EmptyCartAsync(string userId)
         {
             Console.WriteLine($"EmptyCartAsync called with userId={userId}");
-            userCartItems[userId] = emptyCart;
+            userCartItems[userId] = new Hipstershop.Cart();
 
             return Task.CompletedTask;
         }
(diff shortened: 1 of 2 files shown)


src/cartservice/cartstore/LocalCartStore.cs
```diff
@@ -65,7 +65,7 @@ public Task AddItemAsync(string userId, string productId, int quantity)
         public Task EmptyCartAsync(string userId)
         {
             Console.WriteLine($"EmptyCartAsync called with userId={userId}");
-            userCartItems[userId] = emptyCart;
+            userCartItems[userId] = new Hipstershop.Cart();
```
DanSanche: It looks like we can remove the emptyCart variable on line 28, right?

askmeegs: It's used here -- https://github.com/GoogleCloudPlatform/microservices-demo/blob/e5fb708cf2e5ae9d8a5b71746b1868c6fe9ead8c/src/cartservice/cartstore/LocalCartStore.cs#L80

DanSanche: I'm curious whether it would make more sense to just instantiate a new cart there as well. But I'm not really familiar enough with the language to say

PR #293, opened 2020-02-15
Adds option to disable Stackdriver Export

askmeegs:
Adds environment variable option to disable Stackdriver export for Metrics and Tracing, along with the option to turn off Stackdriver Profiler and Debugger. 

All services still default to `True` for exporting to SD - must manually add the env variables to disable export.

Example deployment YAML: 

```
        env:
        - name: PORT
          value: "7000"
        - name: DEBUGGER
          value: "False"
        - name: TRACE
          value: "False"
        - name: PROFILER
          value: "False"
```

```
shippingservice-d6567fcb7-9dg9z server {"message":"Debugger disabled.","severity":"info","timestamp":"2020-02-15T22:11:30.628788162Z"}
shippingservice-d6567fcb7-9dg9z server {"message":"Tracing disabled.","severity":"info","timestamp":"2020-02-15T22:11:30.628900674Z"}
shippingservice-d6567fcb7-9dg9z server {"message":"Stats disabled.","severity":"info","timestamp":"2020-02-15T22:11:30.629103694Z"}
shippingservice-d6567fcb7-9dg9z server {"message":"Shipping Service listening on port :50051","severity":"info","timestamp":"2020-02-15T22:11:30.629151017Z"}
```


ahmetb:
I think my main feedback would be: Do not parse booleans.
YAML especially is notoriously bad for storing booleans.

Instead of doing TRACE="False", consider checking for existence of a flag, for example `DISABLE_TRACE`, and any value that's set (even `0`) would just disable it.

ahmetb:
^ This alone would simplify a lot of the current code.

src/adservice/src/main/java/hipstershop/AdService.java
```diff
@@ -213,21 +213,21 @@ private void blockUntilShutdown() throws InterruptedException {
         .build();
   }
 
-  private static void initStackdriver() {
-    logger.info("Initialize Stackdriver");
-
+  private static void initStats() {
+    String statsEnv = System.getenv("STATS");
+    if (statsEnv == null) {
+      statsEnv = "true";
```
ahmetb: this is kinda nasty. especially because 1) it's a string 2) the yaml has capitalized values like "False".

src/shippingservice/main.go
```diff
@@ -58,8 +59,36 @@ func init() {
...
+	if err != nil {
+		log.Errorf("Could not get TRACE var: %v, defaulting to True", err)
+		trace = true
+	}
+
+	profiler, err := getenvBool("PROFILER")
+	if err != nil {
+		log.Errorf("Could not get PROFILER var: %v, defaulting to True", err)
+		profiler = true
+	}
```
ahmetb: imagine just doing

    profilerEnabled := os.Getenv("DISABLE_PROFILER") == ""

no parsing methods, no errors, no error handling, no boolstrings in yaml

askmeegs: thanks for this! you're totally right, for some reason I thought it was important to check true/false, but simply checking if the env is set is a much better option. updated all the services to remove bool parsing.

askmeegs:
@ahmetb this look ok to merge?

kubernetes-manifests/adservice.yaml
```diff
@@ -34,6 +34,10 @@ spec:
         env:
         - name: PORT
           value: "9555"
+        # - name: DISABLE_STATS
+        #   value: 1
```
ahmetb: sadly these `1`s in YAML need to be `"1"` otherwise the validation will fail AFAIK.

kubernetes-manifests/emailservice.yaml
```diff
@@ -34,8 +34,10 @@ spec:
         env:
         - name: PORT
           value: "8080"
-        - name: ENABLE_PROFILER
```
ahmetb: if the default in the release was "profiler disabled", might be wise to uncomment that in the final result.

ahmetb:
Couple nits, otherwise looking good.

kubernetes-manifests/adservice.yaml
askmeegs: Hmm ok, I think I was able to "kubectl apply" these, but quoted them anyway.

Issue #292, opened 2020-02-12
adservice image faild if you want to build from root directory


danielyaghini:
in the adservice folder would be described:
to build the docker image
From the repository root, run:
docker build --file src/adservice/Dockerfile .

but it caus the following error:
Step 3/17 : COPY ["build.gradle", "gradlew", "./"]
COPY failed: stat /var/lib/docker/tmp/docker-builder288251425/build.gradle: no such file or directory

ahmetb:
What OS are you building on?

danielyaghini:
with Github action runs-on: ubuntu-latest

ahmetb:
@DanSanche has our GH Workflows worked so far? It seems like the hosted worker is not working.

danielyaghini:
hier my workflow unter github action:
`
name: Docker Image CI
on: [push]
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2
    - name: Build the Docker image
      run: docker build --file src/adservice/Dockerfile .`

can you see if you can reproduce the error?

danielyaghini:
Here is my workflow under the github action:
`
name: Docker Image CI

on: [push]

jobs:
  build:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2
    - name: Build the Docker image
      run: docker build --file src/adservice/Dockerfile .
`

cloud you check if you can reproduce the error?

DanSanche:
@ahmetb GitHub actions haven't been enabled for the repo yet. This is separate from that branch

It looks like the adservice README instructs you to build the container from the repo root, but the Dockerfile expects you to build it from the src/adservice directory. Is there any reason we'd want to build the Dockerfile from the root? That seems unnecessarily. I think we should just update the README

@danielyaghini it should work if you just use `docker build --file src/adservice/Dockerfile src/adservice` instead

DanSanche:
closing this since we updated the README to build from `src/adservice`

PR #296, opened 2020-02-21
Update container build instructions

DanSanche:
Update the instructions for how to build docker image in services with misleading READMEs

addresses https://github.com/GoogleCloudPlatform/microservices-demo/issues/292

### DIFF
--- a/src/adservice/README.md
+++ b/src/adservice/README.md
@@ -20,9 +20,9 @@ If you need to upgrade the version of gradle then run
 
 ## Building docker image
 
-From the repository root, run:
+From `src/adservice/`, run:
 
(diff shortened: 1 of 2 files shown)


ahmetb:
This would also work

From the repository root, run “docker build -t foo src/adservice”

DanSanche:
Would you prefer that way? I was thinking since they're reading the README in the src/X folder, it makes sense to give build directions relative to src/X. But I'm happy with either way

ahmetb:
I don't have any preference, really.  These individual readmes are probably also not necessary if there aren't any non-trivial info in them.

PR #319, opened 2020-04-08
Improve smoke test

DanSanche:
Smoke test would previously just check for a 200 status at http://frontend. This would miss serious errors in other services

Now, the test checks the output of loadgenerator for errors, ensuring a number of endpoints are queried without fail. This should be a better metric to ensure the app as a whole is healthy

Note: This is expected to fail until https://github.com/GoogleCloudPlatform/microservices-demo/pull/318 is merged

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,8 @@ jobs:
         kubectl get nodes
     - name: Deploy From Source
       run: |
-        skaffold run
+        skaffold config set --global local-cluster true
+        skaffold run --default-repo local
     - name: Wait For Pods
       timeout-minutes: 20
       run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,10 +48,10 @@ jobs:
         REQUEST_COUNT="0"
         while [[ "$REQUEST_COUNT"  -lt "50"  ]]; do
             sleep 5
-            REQUEST_COUNT=$(kubectl logs -l app=loadgenerator | grep Total | awk '{print $2}')
+            REQUEST_COUNT=$(kubectl logs -l app=loadgenerator | grep Aggregated | awk '{print $2}')
         done
         # ensure there are no errors hitting endpoints
-        ERROR_COUNT=$(kubectl logs -l app=loadgenerator | grep Total | awk '{print $3}' | sed "s/[(][^)]*[)]//g")
+        ERROR_COUNT=$(kubectl logs -l app=loadgenerator | grep Aggregated | awk '{print $3}' | sed "s/[(][^)]*[)]//g")
         if [[ "$ERROR_COUNT" -gt "0" ]]; then
           exit 1
         fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,12 +42,16 @@ jobs:
       timeout-minutes: 5
       run: |
         set -x
-        RESULT=" "
-        while [[ "$RESULT" != "  HTTP/1.1 200 OK" ]]; do
-          sleep 1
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -42,12 +42,16 @@ jobs:
...
+        kubectl delete pod -l app=loadgenerator
+        # wait for requests to come in
+        REQUEST_COUNT="0"
+        while [[ "$REQUEST_COUNT"  -lt "50"  ]]; do
+            sleep 5
+            REQUEST_COUNT=$(kubectl logs -l app=loadgenerator | grep Aggregated | awk '{print $2}')
         done
-        if [[ "$RESULT" != "  HTTP/1.1 200 OK" ]]; then 
+        # ensure there are no errors hitting endpoints
+        ERROR_COUNT=$(kubectl logs -l app=loadgenerator | grep Aggregated | awk '{print $3}' | sed "s/[(][^)]*[)]//g")
```
murog, line 54: a comment about what the sed string does could be helpful (-:

.github/workflows/ci.yml
```diff
@@ -42,12 +43,16 @@ jobs:
...
+        if [[ "$ERROR_COUNT" -gt "0" ]]; then
```
ahmetb, line 56: I feel like startup errors or small flakes can instantly fail the smoke test. I wish there was a better way though.

DanSanche: Yeah, hopefully over time we can add more robust tests. For now it's easy enough to re-run though (start up errors are reduced by re-starting just the loadgenerator at the start of the test. It's usually pretty stable)

.github/workflows/ci.yml
ahmetb: yeah, similarly I'm not sure which part of the command actually ensures this matches only 1 occurrence of `Aggregated`.

ahmetb: For example my `kubectl logs -l app=loadgenerator -c main` on cluster `demo-app` on project `microservices-demo-app` right now returns:



```
$ kubectl logs -l app=loadgenerator -c main
 GET /product/66VCHSJNUP                                       119217    39(0.03%)      88      25    2735  |      40    0.30
 GET /product/6E92ZMYYFZ                                       119973    37(0.03%)      87      27    3360  |      40    0.20
 GET /product/9SIQT8TOJO                                       119565    33(0.03%)      87      27    3230  |      40    0.20
 GET /product/L9ECAV7KIM                                       119289    24(0.02%)      88      27    2828  |      39    0.10
 GET /product/LS4PSXUNUM                                       119623    36(0.03%)      88      26    2995  |      39    0.00
 GET /product/OLJCESPC7Z                                       119846    44(0.04%)      89      26    2970  |      40    0.00
 POST /setCurrency                                             165396    47(0.03%)     114      36    3094  |      69    0.20
--------------------------------------------------------------------------------------------------------------------------------------------
 Total                                                        1902768   558(0.03%)                                       1.60
```

So this command above returns 0, despite 558 failures.

If I actually tail with `-f` I get more lines.

Even when loadgenerator prints its output every few secs once I do `kubectl logs -f`, I still don't see word `Aggregated`. I assume it's printed at the beginning, then lost.

DanSanche: @ahmetb it looks like you're using an older version of locust. It was updated to 0.13 [here](https://github.com/GoogleCloudPlatform/microservices-demo/commit/997b5e88072d0948a93b40ccf8c27eb1fcb66ebe#diff-dd74128e3994eeda8fd49072b5901c42), which uses [Aggregated instead of Total](https://docs.locust.io/en/stable/changelog.html#id8)

ahmetb: ah yeah, thanks. I guess I forgot we don't auto-deploy the new versions to the demo cluster.

PR #332, opened 2020-04-25
Handle error template errors

bvwells:


bvwells:
Thanks @Daniel-Sanche!

PR #357, opened 2020-06-04
Updated broken loadgenerator to support locust 1.0.2

adrienaflalo:
Locust 1.0 release introduced breaking changes such as class and args renaming. Here are updated version of loadconcentrator to support latest locust version


src/loadgenerator/loadgen.sh
```diff
@@ -32,4 +32,4 @@ if test $STATUSCODE -ne 200; then
 fi
 
 # else, run loadgen
-locust --host="http://${FRONTEND_ADDR}" --no-web -c "${USERS:-10}" 2>&1
```
ahmetb, line 35: can we omit -r? It's not defined, so adding a default -r=10 potentially changes behavior (that's not something we want).

adrienaflalo: Done

djmailhot (approved):
Approved.

CI workflows failed due to bad GitHub runner config.  Ran the code locally, works as intended.

PR #364, opened 2020-06-30
Use buildkit by default

inductor:
This repo's base images use multi stage build and I think it should be better to make the build faster by BuildKit.

### DIFF
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -17,7 +17,7 @@ kind: Config
 build:
   artifacts:
   # image tags are relative; to specify an image repo (e.g. GCR), you
-  # must provide a "default repo" using one of the methods described 
+  # must provide a "default repo" using one of the methods described
   # here:
   # https://skaffold.dev/docs/concepts/#image-repository-handling
   - image: emailservice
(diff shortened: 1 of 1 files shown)


ahmetb:
I'm not sure about the compatibility outcomes of this.

inductor:
@ahmetb 
Based on the Skaffold source code, it adds the environment variable.
https://github.com/GoogleContainerTools/skaffold/blob/17e1e021b7e68ee57c717c1c22e06e48c68403a9/pkg/skaffold/build/local/docker.go#L89

```go
	if b.cfg.UseBuildkit {
		cmd.Env = append(cmd.Env, "DOCKER_BUILDKIT=1")
	}
```

It does not break anything where you don't have BuildKit capability on your Docker host.

ahmetb:
Makes me wonder, why then the Skaffold doesn't do it by default?

inductor:
Um I think that is because Docker hasn't set it by default?

djmailhot (approved):
Seems to improve build time by 50% (from ~3m to ~2m).  Promises only to be run if buildkit is available.

Passes CI tests on internal branch.

PR #367, opened 2020-07-13
Add licenses

Daniel-Sanche:
Google OSS projects are meant to have Google license headers in every file. This PR adds the missing headers, and adds a test to the CI system to catch future issues


.github/workflows/ci.yml
```diff
@@ -14,6 +28,16 @@ jobs:
     runs-on: self-hosted
     steps:
     - uses: actions/checkout@v2
+    - name: License Check
+      run: |
+        set -x
+        git init
+        git add --all
+        git -c user.name="CI Bot" -c user.email="<>" commit -m "initial state"
```
ahmetb, line 36: does `addlicense` need to make a commit? is that why you need this?

.github/workflows/ci.yml
```diff
@@ -14,6 +28,16 @@ jobs:
...
+        addlicense ./
+        if [[ -n $(git status -s) ]]; then
```
ahmetb, line 38: you can also do simply do this: `test -z "$(git status -s | tee /dev/stderr)"`

right now you're not printing which files have a diff.

.github/workflows/ci.yml
Daniel-Sanche: It doesn't need to commit, just see the status. 

I wrote this check a while ago for Bank of Anthos. I assume I got an error at some point when this was left out, but I can double check to see if it still works with this line removed

Daniel-Sanche: Actually, I remember the context now: there is no git repo in the filesystem here originally, so I needed to create one and make an initial commit as a way of checking for changes after running the `addlicence` command.

It's slightly hacky, but this was the cleanest solution I could think of. If you have other ideas for how to compare file system states, I'm open to other ideas

.github/workflows/ci.yml
Daniel-Sanche: Actually the results are printed, since the `set -x` flag is on. I think `git status -s` is preferable to `git diff` because we want to see the list of the files, not the full license content for each file

ahmetb: sounds good. you can still remove this in this case by doing:

git init && git add --all && addlicense . && test -z $(**git diff --name-only** | tee /dev/stderr)

Daniel-Sanche: Personally I find that a bit less readable, but I made the change

PR #369, opened 2020-07-15
fix: update locust files for current locust version

jorgonzalez:
Current locust configuration doesn't work with stable locust version from pip install.

### DIFF
--- a/src/loadgenerator/loadgen.sh
+++ b/src/loadgenerator/loadgen.sh
@@ -32,4 +32,4 @@ if test $STATUSCODE -ne 200; then
 fi
 
 # else, run loadgen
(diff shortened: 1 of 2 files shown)


jorgonzalez:
> We keep changing locust more than we update the project itself.
> 
> [35edb74](https://github.com/GoogleCloudPlatform/microservices-demo/commit/35edb7467857543367a36c9e680ed533c5b6e3f2) is the offending commit. Technically this is locust project's problem that they keep breaking users between patch releases.
> 
> It's a better investment to pin to a particular version and never update it until the package gets deleted.

You are right, the previous setup worked and had correctly pinned versions.

PR #378, opened 2020-08-05
Fix multiply item price by quantity

keziasulami:
Bug in "Total Paid" after checkout with quantity more than 1.


djmailhot (approved):
Approved.  Good catch!

Deployed and tested manually.  Verified to work as intended.

PR #384, opened 2020-08-25
Use pip prefix option

matrohon:
Using pip --install-option leads to the following error when building the loadgenerator image:

```
ERROR: Location-changing options found in --install-option: ['--prefix'] from command line. This is unsupported, use pip-level options like --user, --prefix, --root, and --target instead.
```

The pip prefix option must be used instead.


ahmetb:
This is fixed in another PR.

PR #398, opened 2020-09-10
Fix sfx config

owais:


ahmetb:
Do you mind explaining why/how you ended up accidentally making a PR to the upstream?

owais:
Unintentionally :)

owais:
I switched the repo I wanted to created against and then clicked the submit button without realizing that changing the dropdown from repo reloads the whole page and I needed to wait for that. :)

owais:
Looks like there is no way to configure a fork to default all PRs to the fork instead of upstream. I'll talk to the owner and see if we can move the repo to an indepenent one instead of a fork.  That should prevent future accidents. That said, it would have been nice if Github fixed the UX pattern in the first place

PR #402, opened 2020-09-15
Swap in a published test credit card number

jmound:
Is it possible to swap out the current `4432-8015-6152-0454` in exchange for a published test card number? `4242-4242-4242-4242` was obtained from https://stripe.com/docs/testing#cards


ahmetb:
You can fork and do that. 
I don't understand the rationale behind the change.

jmound:
Understood, and no problem. Thanks for the quick feedback.

PR #410, opened 2020-10-13
Move to go modules

### LINKED ISSUE #408: Migrate to Go modules
Right now we're using `dep` to manage dependencies for Go services.
On clean builds without cache (e.g. Google Cloud Builds), Go services are the slowest building ones among all (to my surprise).

Migrating to go.mod can help with the build speed significantly as most of the time is spent downloading and unpacking dependencies.
This was also brought up in #295.

Most notably, this dependency number 21 is exactly where the build gets stuck for a long time (I am assuming it's some sort of unarchival step, or maybe FS is really slow somehow):

image

ahmetb:
Simplifies and modernizes the go build process.
Old vs new build times (without cache):

- frontend: 185s vs 33s.
- productcatalog: 190s vs 30s.
- shippingservice: 200s vs 36s.
- checkoutservice: 129s vs 27s.

Also upgrades images to go1.15.
Fixes #408.


djmailhot (approved):
LGTM

ahmetb:
If anything about Go services suddenly breaks in the next few days this is probably why. :)

PR #411, opened 2020-10-13
Add rollbar configuration to loadgenerator

letaniaferreira:
Add rollbar configuration to the loadgenerator microservice


ahmetb:
What does it do and why do we need it? :)

PR #412, opened 2020-10-13
Fix commands to set up GitHub actions runner

djmailhot:
Add `git` installation
Add `-yq` flags to apt-get commands
Always use `apt-get` instead of `apt`.


ahmetb (commented):
might as well add -qq instead of single -q

`-qq No output except for errors`

djmailhot:
@ahmetb could use another review on this PR, thanks!

ahmetb (approved):

PR #413, opened 2020-10-13
Update loadgenerator Dockerfile to fix Pip incompatibility

### LINKED ISSUE #406: failed to build: couldn't build "loadgenerator"
- Precisely:
ERROR: Failed building wheel for gevent

-  Solution: change in `src/loadgenerator/Dockerfile`
FROM python:3-slim as base    -->    FROM python:3.6-slim as base

djmailhot:
Resolves #406

### DIFF
--- a/src/loadgenerator/Dockerfile
+++ b/src/loadgenerator/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM python:3.7-slim as base
+FROM python:3.9-slim as base
 
 FROM base as builder
 
--- a/src/loadgenerator/Dockerfile
+++ b/src/loadgenerator/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM python:3.9-slim as base
+FROM python:3.8-slim as base
 
 FROM base as builder
 
--- a/src/loadgenerator/Dockerfile
+++ b/src/loadgenerator/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM python:3.7.2-slim as base
+FROM python:3-slim as base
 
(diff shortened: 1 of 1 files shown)


src/loadgenerator/Dockerfile
```diff
@@ -12,16 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM python:3-slim as base
+FROM python:3.7.2-slim as base
```
ahmetb, line 15: refrain from using specific versions like 3.7.2.
if there's a security bug tomorrow, 3.7.2 will potentially be deleted.

src/loadgenerator/Dockerfile
```diff
@@ -12,16 +12,17 @@
...
 FROM base as builder
 
 RUN apt-get -qq update \
     && apt-get install -y --no-install-recommends \
-        g++
+        g++ libzmq3-dev
 
 COPY requirements.txt .
 
+RUN pip install --upgrade pip
```
ahmetb, line 25: is this actually required? 
if you use a sufficiently high version of python:3.x you should be getting a good pip.

in fact this goes against "reproducible builds" mantra, because you're installing "whatever is the most recent pip available" in this command ––which is not a good idea for reproducibility.

ahmetb (approved):

src/loadgenerator/Dockerfile
djmailhot: Done

src/loadgenerator/Dockerfile
djmailhot: Done

djmailhot:
@ahmetb 
To get the CI tests to pass, I've found that in need to pin the Docker image to an older python version
`python3-slim` results in a build failure for a dependency of `locust` the load generation package.
`python3.8-slim` results in a successful build, and thus the CI tests pass.

Currently, every CI run for every PR is failing because the code currently uses `python3-slim`.

Issue #431, opened 2020-11-08
failed to build paymentservice, failed to build LLB, failed to load cache key


npitaev:
Facing build issue for the paymentservice, which might be similar to the issue #406 failed to build: couldn't build "loadgenerator", however the solution from that issue, does not help. IMO we need somehow to tweak src/paymentservice/Dockerfile but I am not sure how.

Error while following option 2 install on GCP - the following output after executing skaffold run --default-repo=gcr.io/[my valid project ID]

Building [paymentservice]...
[+] Building 0.3s (5/13)
 => [internal] load build definition from Dockerfile                     0.0s
 => => transferring dockerfile: 1.32kB                                   0.0s
 => [internal] load .dockerignore                                        0.0s
 => => transferring context: 53B                                         0.0s
 **=> ERROR [internal] load metadata for docker.io/library/node:8-alpine   0.1s
 => ERROR [base 1/1] FROM docker.io/library/node:8-alpine                0.1s**
 => => resolve docker.io/library/node:8-alpine                           0.1s
 => [internal] load build context                                        0.0s
 => => transferring context: 145.14kB                                    0.0s

  [internal] load metadata for docker.io/library/node:8-alpine:

  [base 1/1] FROM docker.io/library/node:8-alpine:

f**ailed to solve with frontend dockerfile.v0: failed to build LLB: failed to load cache key: docker.io/library/node:8-alpine not found
failed to build: couldn't build "paymentservice": running build: exit status 1**

askmeegs:
Thanks for reporting this. I'm investigating.
I'm also seeing this error in cartservice: 

```
Building [cartservice]...
[+] Building 0.5s (7/14)
 => [internal] load build definition from Dockerfile                                                                                                                                                             0.0s
 => => transferring dockerfile: 1.38kB                                                                                                                                                                           0.0s
 => [internal] load .dockerignore                                                                                                                                                                                0.0s
 => => transferring context: 64B                                                                                                                                                                                 0.0s
 => [internal] load metadata for docker.io/library/alpine:3.8                                                                                                                                                    0.3s
 => ERROR [internal] load metadata for docker.io/microsoft/dotnet:2.1-sdk-alpine                                                                                                                                 0.1s
 => ERROR [builder 1/4] FROM docker.io/microsoft/dotnet:2.1-sdk-alpine                                                                                                                                           0.1s
[...]

PR #436, opened 2020-11-10
skaffold - Turn off Docker buildkit by default. 

askmeegs:
Addresses #431. Enabling buildkit causes skaffold user errors - root cause not yet identified. This is a patch PR to revert #364 to allow builds to run again.


djmailhot (approved):
LGTM

PR #439, opened 2020-11-11
fixed cnd url for bootstrap files and upgrade 4.1.1 -> 4.1.3

fmarinelli:
Fixes # .

Change summary:
- 

Remaining issues / concerns:

### DIFF
--- a/src/frontend/templates/footer.html
+++ b/src/frontend/templates/footer.html
@@ -30,8 +30,7 @@
         </div>
     </div>
 </footer>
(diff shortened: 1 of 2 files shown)


ahmetb:
Why the CDN URL change? 
We're striving for external deps that sustain long term. It seems https://www.jsdelivr.com/ is maintained on personal capacity. So is https://www.bootstrapcdn.com/. I recommend using https://cdnjs.com/ if possible, it's a tad bit more official and likely to last several years.

fmarinelli:
You are right, I will change back to bootstrapcdn and redo the PR

PR #481, opened 2021-01-02
minor updates for cartservice

mathieu-benoit:
- `ASPNETCORE_URLS` env in `Dockerfile`
- C# 9 syntax improvement with `Program.cs`
- `Grpc.AspNetCore` - 2.34.0
- `GRPC_HEALTH_PROBE_VERSION` - `v0.3.6`
- Remove unused files + enhance `.dockerignore` file


src/cartservice/Dockerfile
```diff
@@ -22,9 +22,10 @@ RUN dotnet publish cartservice.csproj -p:PublishSingleFile=true -r linux-musl-x6
 
 # https://mcr.microsoft.com/v2/dotnet/runtime-deps/tags/list
 FROM mcr.microsoft.com/dotnet/runtime-deps:5.0.1-alpine3.12-amd64
-RUN GRPC_HEALTH_PROBE_VERSION=v0.3.5 && \
+RUN GRPC_HEALTH_PROBE_VERSION=v0.3.6 && \
     wget -qO/bin/grpc_health_probe https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/${GRPC_HEALTH_PROBE_VERSION}/grpc_health_probe-linux-amd64 && \
     chmod +x /bin/grpc_health_probe
 WORKDIR /app
 COPY --from=builder /cartservice .
+ENV ASPNETCORE_URLS http://*:7070
```
askmeegs, line 30: just curious, what is the functionality of this? was cartservice using a .NET default, prior to this?

askmeegs (approved):

src/cartservice/Dockerfile
mathieu-benoit: Sure, this env var is one way to set the URL, especially when not using the default port: https://andrewlock.net/5-ways-to-set-the-urls-for-an-aspnetcore-app/. Before this PR, this env var was passed in the Kubernetes deployment file, now it's embedded in the Dockerfile instead. Hope that makes sense?

askmeegs: Yes! That clarifies, thanks!

PR #507, opened 2021-02-17
Made Loadgenerator capable of sending calls through https and outside the cluster

knee-berts:
Fixes # .

Change summary:
Adjusted loadgen shell script to include the entire http call so that the envvar can be adjusted for https calls. This allows for using the loadgenerator to call an external https lb.
- 

Remaining issues / concerns:

### DIFF
--- a/kubernetes-manifests/loadgenerator.yaml
+++ b/kubernetes-manifests/loadgenerator.yaml
@@ -35,7 +35,7 @@ spec:
         image: loadgenerator
         env:
         - name: FRONTEND_ADDR
-          value: "frontend:80"
+          value: "http://frontend:80"
         - name: USERS
           value: "10"
         resources:
--- a/src/loadgenerator/loadgen.sh
+++ b/src/loadgenerator/loadgen.sh
@@ -25,11 +25,11 @@ fi
 set -x
 
 # if one request to the frontend fails, then exit
-STATUSCODE=$(curl --silent --output /dev/stderr --write-out "%{http_code}" http://${FRONTEND_ADDR})
+STATUSCODE=$(curl --silent --output /dev/stderr --write-out "%{http_code}" ${FRONTEND_ADDR})
 if test $STATUSCODE -ne 200; then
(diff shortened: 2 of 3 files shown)


kubernetes-manifests/loadgenerator.yaml
```diff
@@ -35,7 +35,7 @@ spec:
         image: loadgenerator
         env:
         - name: FRONTEND_ADDR
-          value: "frontend:80"
+          value: "http://frontend:80"
```
askmeegs, line 38: Hi Nicholas, could you give some more details on why you added this? This default value, `frontend`, is a Kubernetes service name. it is intended to be reached inside the cluster, and expands out to the kubedns name `frontend.<namespaces>.svc.cluster.local`. I'm not sure what adding an `http` is intended to do here.

src/loadgenerator/loadgen.sh
```diff
@@ -25,11 +25,11 @@ fi
 set -x
 
 # if one request to the frontend fails, then exit
-STATUSCODE=$(curl --silent --output /dev/stderr --write-out "%{http_code}" http://${FRONTEND_ADDR})
+STATUSCODE=$(curl --silent --output /dev/stderr --write-out "%{http_code}" ${FRONTEND_ADDR})
```
askmeegs, line 28: Same here - could you please give more details on your use case? Also why can't a user just add `http://<my-custom-addr>:<port>` to the loadgen deployment manifest, as `FRONTEND_ADDR`? why is an extra http needed here? in this PR as is, wouldn't the curl command actually have two `http://http://{FRONTEND_ADDR}`? because it's here, and in the deployment manifest?

askmeegs (changes requested):
Hi Nicholas, a few questions.

knee-berts:
I made a mistake on changing the envvar, but what would be great is to include the protocol in the envvar string. Otherwise you won't be able to address https as it is hard coded. Let me fix and repull.

PR #512, opened 2021-03-18
Disable the PROFILER and DEBUGGER by default

yongfengdu:
DIsable the Profiler and Debugger to avoid CrashLoop while run the demo outside of GCP.
Otherwise, the recommendationservice will fail health check before 8080 port ready.

Fixes # .

Change summary:
- 

Remaining issues / concerns:

### DIFF
--- a/kubernetes-manifests/recommendationservice.yaml
+++ b/kubernetes-manifests/recommendationservice.yaml
@@ -47,10 +47,10 @@ spec:
           value: "productcatalogservice:3550"
         # - name: DISABLE_TRACING
         #   value: "1"
-        # - name: DISABLE_PROFILER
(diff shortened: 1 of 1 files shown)


askmeegs:
Hello, 

We provide the `DISABLE_PROFILER` and `DISABLE_DEBUGGER` env variables so that users can set to `1` locally / disabling export in non GCP environments. You can change these locally to prevent a crash loop in a non GCP cluster, but because this is a GCP (GKE) demo, we will keep these fields enabled by default. 

If you continue to encounter issues with this or other areas of the app, please file a Github issue here:  https://github.com/GoogleCloudPlatform/microservices-demo/issues 

Thanks!

PR #516, opened 2021-03-23
Fix dotnet CLI tests for Github actions

### LINKED ISSUE #428: Fix dotnet CLI tests for Github actions
Currently failing with a relative path error, commented out in the workflow YAMLs right now.

deesejohn:
Fixes #428

Change summary:
- Created `src` directory for api project
- Created solution file to relate api and unit test projects
- Refactored unit tests to use TestHost
- Uncommented github workflow for C# Unit Tests
- Added `actions/setup-dotnet@v1` to github workflows to support dotnet v5.0.201


deesejohn:
@googlebot I signed it!

.github/workflows/ci-master.yaml
```diff
@@ -24,6 +24,9 @@ jobs:
     runs-on: self-hosted
     steps:
     - uses: actions/checkout@v2
+    - uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: '5.0.201'
```
askmeegs, line 29: Quick question - without this step, would your C# tests not run? Or does this just explicitly define the version needed?

src/cartservice/src/appsettings.json
```diff
@@ -11,5 +11,6 @@
       "EndpointDefaults": {
         "Protocols": "Http2"
       }
-    }
+    },
+    "Urls": "http://*:7070"
```
askmeegs, line 15: Where is this used?  is it required?

src/cartservice/tests/CartServiceTests.cs
```diff
@@ -90,20 +110,24 @@ public async Task AddItem_ItemExists_Updated()
...
-            await client.EmptyCartAsync(new EmptyCartRequest{ UserId = userId });
+            await client.EmptyCartAsync(new EmptyCartRequest { UserId = userId });
         }
 
         [Fact]
         public async Task AddItem_New_Inserted()
         {
-            string userId = Guid.NewGuid().ToString();
+            // Setup test server and client
+            using var server = await _host.StartAsync();
```
askmeegs, line 120: I notice a few lines here (the server/client setup) are duplicated between tests, do you think it's worth writing a helper function or okay to leave as-is?

askmeegs (commented):
Overall looks great! just a few clarifying questions, otherwise LGTM.

.github/workflows/ci-master.yaml
deesejohn: The runtime was defaulting to 3.1 and needed to be set explicitly for the tests to pass. Maybe the self-hosted instance simply needs the latest runtime?

src/cartservice/src/appsettings.json
deesejohn: This is not required and was simply added when I was running the application natively and needed to replicate the port set within the Dockerfile here: https://github.com/GoogleCloudPlatform/microservices-demo/blob/a7ad8f7cc7cd15e7db024c7b554b04e0d1fbbe52/src/cartservice/Dockerfile#L32

I will revert this change.

src/cartservice/tests/CartServiceTests.cs
deesejohn: I would suggest keeping it as-is to keep top-to-bottom readability clear. I'd be happy migrate to a private method if preferred.

.github/workflows/ci-master.yaml
askmeegs: Thanks!

src/cartservice/tests/CartServiceTests.cs
askmeegs: Sounds good!

askmeegs (approved):
Verified that dotnet tests pass locally. Thanks for contributing this work @deesejohn!

PR #529, opened 2021-04-07
Set ENV_PLATFORM default to `local`, auto-detect `gcp`

### LINKED ISSUE #334: Set ENV_PLATFORM default at runtime
Currently, the frontend's `ENV_PLATFORM` flag defaults to `gcp` and depends on the user setting an env variable.

Instead, we should dynamically se `ENV_PLATFORM`. Default should be `local`, and set to `gcp` only if `net.LookupHost("metadata.google.internal.")` returns no error. 

Other platforms (`aws`, `azure`) can still be controlled with the env variable.

askmeegs:
Fixes #334 

Tested on:
- GKE (ENV_PLATFORM set to local, sees metadata server, correctly sets flag to GCP) 
- Minikube with `local` flag set --> sets to local 
- Minikube with no flag set  --> sets to local 
- Minikube with manual flag override to AWS --> sets to aws

### DIFF
--- a/src/frontend/handlers.go
+++ b/src/frontend/handlers.go
@@ -47,6 +47,8 @@ var (
 	plat platformDetails
 )
 
+var validEnvs = []string{"local", "gcp", "azure", "aws", "onprem"}
+
 func (fe *frontendServer) homeHandler(w http.ResponseWriter, r *http.Request) {
 	log := r.Context().Value(ctxKeyLog{}).(logrus.FieldLogger)
 	log.WithField("currency", currentCurrency(r)).Info("home")
@@ -82,9 +84,12 @@ func (fe *frontendServer) homeHandler(w http.ResponseWriter, r *http.Request) {
 
 	// Set ENV_PLATFORM (default to local if not set; use env var if set; otherwise detect GCP, which overrides env)_
 	var env = os.Getenv("ENV_PLATFORM")
-	if env == "" {
+	// Only override from env variable if set + valid env
+	if env == "" || stringinSlice(validEnvs, env) == false {
+		fmt.Println("env platform is either empty or invalid")
 		env = "local"
 	}
+	// Autodetect GCP
 	addrs, err := net.LookupHost("metadata.google.internal.")
 	if err == nil && len(addrs) >= 0 {
 		log.Debugf("Detected Google metadata server: %v, setting ENV_PLATFORM to GCP.", addrs)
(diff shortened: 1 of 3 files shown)


src/frontend/handlers.go
```diff
@@ -79,8 +80,20 @@ func (fe *frontendServer) homeHandler(w http.ResponseWriter, r *http.Request) {
...
-	var env = os.Getenv("ENV_PLATFORM")
+	// Set ENV_PLATFORM (checks for GCP. otherwise uses env value.)
+	var env = "local"
+	addrs, err := net.LookupHost("metadata.google.internal.")
+	if err == nil && len(addrs) >= 0 {
+		log.Debugf("Detected Google metadata servers: %v, setting ENV_PLATFORM to GCP.", addrs)
+		env = "gcp"
+	} else {
+		var envFromConfig = os.Getenv("ENV_PLATFORM")
+		fmt.Printf("⭐️ Env from config is: %s", envFromConfig)
```
bourgeoisor, line 91: Do we have other emoji's in print statements? I love emoji! Just making sure we stay consistent. 😄

src/frontend/handlers.go
```diff
@@ -79,8 +80,20 @@ func (fe *frontendServer) homeHandler(w http.ResponseWriter, r *http.Request) {
 		ps[i] = productView{p, price}
 	}
 
-	//get env and render correct platform banner.
-	var env = os.Getenv("ENV_PLATFORM")
+	// Set ENV_PLATFORM (checks for GCP. otherwise uses env value.)
+	var env = "local"
+	addrs, err := net.LookupHost("metadata.google.internal.")
```
bourgeoisor, line 85: I wonder if there's a way to do this without hard-coding a conditional for it. How is for example the Azure case handled? Do we always expect the user / something to set the env correctly, which is why we don't have to handle it here? Does it make sense to do something similar for GCP, or are we doing it more explicitly to remove a manual step somewhere else?

src/frontend/handlers.go
askmeegs: nope that was just for debugging, good catch! removing.

src/frontend/handlers.go
```diff
@@ -79,8 +80,20 @@ func (fe *frontendServer) homeHandler(w http.ResponseWriter, r *http.Request) {
 		ps[i] = productView{p, price}
 	}
 
-	//get env and render correct platform banner.
-	var env = os.Getenv("ENV_PLATFORM")
+	// Set ENV_PLATFORM (checks for GCP. otherwise uses env value.)
+	var env = "local"
```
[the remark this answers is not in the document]
askmeegs: Right, if the user removes the env var entirely I wanted to avoid a glitchy frontend. I instead commented out the default env in the YAML, and set the local default in the code only.

src/frontend/handlers.go
askmeegs: Good point, I added a validity check for the env var to make sure it's a flag we have CSS for. (the flags each have different colors.)  This PR is really just designed to address the request to autodetect GCP. not sure if there's a way to autodetect azure + aws in the same way.

bourgeoisor (approved):
LGTM

PR #530, opened 2021-04-07
Match currency logo to currency

### LINKED ISSUE #483: Changing logo of currency based on currency selected.
image

Can we change the currency icon after changing currency in nav bar

askmeegs:
Fixes #483

### DIFF
--- a/src/frontend/handlers.go
+++ b/src/frontend/handlers.go
@@ -41,7 +41,8 @@ type platformDetails struct {
 var (
 	templates = template.Must(template.New("").
 			Funcs(template.FuncMap{
-			"renderMoney": renderMoney,
+			"renderMoney":        renderMoney,
+			"renderCurrencyLogo": renderCurrencyLogo,
 		}).ParseGlob("templates/*.html"))
 	plat platformDetails
 )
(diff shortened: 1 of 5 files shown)


src/frontend/templates/header.html
```diff
@@ -42,7 +42,7 @@
                 {{ if $.show_currency }}
                 <div class="h-controls">
                     <div class="h-control">
-                        <img src="/static/icons/Hipster_CurrencyIcon.svg" alt="" class="icon" />
+                        <span class="currencyLogo"> {{ renderCurrencyLogo $.user_currency}}</span>
```
bourgeoisor, line 45: Is there a style guide or linting for these HTML templates, out of curiosity?
```suggestion
                        <span class="currencyLogo">{{ renderCurrencyLogo $.user_currency }}</span>
```

src/frontend/handlers.go
```diff
@@ -455,3 +456,19 @@ func cartSize(c []*pb.CartItem) int {
...
 }
+
+func renderCurrencyLogo(currencyCode string) string {
+	logos := map[string]string{}
+	logos["USD"] = "$"
+	logos["CAD"] = "$"
+	logos["JPY"] = "¥"
+	logos["EUR"] = "€"
+	logos["TRY"] = "₺"
+	logos["GBP"] = "£"
```
bourgeoisor, line 467: Thoughts?
```suggestion
	logos := map[string]string{
		"USD": "$",
		"CAD": "$",
		"JPY": "¥",
		"EUR": "€",
		"TRY": "₺",
		"GBP": "£",
	}
```

bourgeoisor (approved):
LGTM! Two minor suggestions.

src/frontend/templates/header.html
askmeegs: Not that I know of, I typically just reference the godoc https://golang.org/pkg/html/template/

PR #535, opened 2021-04-23
Fix for failed docker image build

### LINKED ISSUE #534: Docker builds with make-docker-images.sh fails
Run ./make-release.sh or just run make-docker-images and the build will fail because PR #516 moved cart service artifacts to one folder below into src. 
The changes to the PR was accounted in skaffold.yaml for local builds but it is not propagated to scripts that make release builds. 

Result -> build failure as script fails to find Dockerfile

sitaramkm:
Signed-off-by: Sitaram IYER <sitaramkm@gmail.com>

Fixes #534 .

Change summary:
- Fix building docker image for cartservice. PR 516 moved cartservice build artifacts to one level down to src directory breaking automated builds.
- Special handling cart service with a if condition and appending src to build dir. 

Remaining issues / concerns:
None afaik.

### DIFF
--- a/hack/make-docker-images.sh
+++ b/hack/make-docker-images.sh
@@ -27,9 +27,15 @@ REPO_PREFIX="${REPO_PREFIX:?REPO_PREFIX env variable must be specified}"
 while IFS= read -d $'\0' -r dir; do
     # build image
     svcname="$(basename "${dir}")"
+    builddir="${dir}"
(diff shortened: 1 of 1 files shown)


bourgeoisor (approved):
Verified locally that this diff fixes the cartservice not building correctly and failing the image building process:
```
~ ./hack/make-docker-images.sh
. . .
. . .
Successfully built and pushed all images.
```

PR #557, opened 2021-06-09
Make loadgenerator debug-able

j-windsor:
Fixes #555.

Change summary:
- Calls locust directly from Dockerfile so skaffold python debugger can attach to the pure python process, make other changes to ensure successful debug out of the box.

PR #568, opened 2021-06-24
Add loadgen initcontainer

askmeegs:
Another attempt to fix the failing smoke tests. We believe that PR #557 caused this issue by removing the retry logic in the loadgen's bash script, and instead calling the locustfile directly from the Dockerfile. This PR reintroduces that retry logic by adding an init container with the same curl command used in the startup bash script.  

Here are the loadgen failures after PR 557 merged --  (12% failure) 

```

 Avg     Min     Max  Median  |   req/s failures/s
loadgenerator-5d54c46688-grg9k main  Aggregated                                                       113   14(12.39%)  |     601      13   15319      22  |    1.30    0.00
```

And here is the output before that PR merged -- (no failures) 

```
loadgenerator-64cb8dc78c-ttcsz main  Aggregated                                                        70     0(0.00%)  |      29      13     115      20  |    2.40    0.00
```

Here is the output for this branch (no failures) 

```
loadgenerator-f76ccc598-8bljt main  Aggregated                                                       179     0(0.00%)  |      25      11     124      18  |    2.40    0.00
```

Still not sure exactly why having the loadgen start up and never fail/retry (as before) causes failures continuously -- it should be that once the frontend starts up, the loadgen errors go away.  But it may be that the smoke test is logging results as the loadgen is pinging a frontend that is not ready (ie. collecting results too soon) and that's what is causing the errors.

### DIFF
--- a/.github/workflows/ci-master.yaml
+++ b/.github/workflows/ci-master.yaml
@@ -66,7 +66,7 @@ jobs:
         EOF
         echo Deploying application
         skaffold config set --global local-cluster false
-        skaffold run --default-repo=gcr.io/$PROJECT_ID/$GITHUB_REF --tag=$GITHUB_SHA --namespace=$NAMESPACE --status-check=false
+        skaffold run --default-repo=gcr.io/$PROJECT_ID/$GITHUB_REF --tag=$GITHUB_SHA --namespace=$NAMESPACE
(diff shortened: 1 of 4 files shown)


Shabirmean (approved):
LGTM!

PR #571, opened 2021-06-30
Use dotnet binary to start cartservice so its debuggable with vsdbg

j-windsor:
Change summary:
Start using `dotnet` instead of standalone so it is easily debuggable with vsdbg.


mathieu-benoit:
Just curious here @j-windsor, should it be a dedicated `Dockerfile.debug` file for example with those changes and allow the debugging feature you are mentioning. This, in order to keep the current one light/small ready for Production/Release mode?

j-windsor:
@mathieu-benoit great idea! I adjusted accordingly.

skaffold.yaml
```diff
@@ -65,3 +67,11 @@ profiles:
       diskSizeGb: 300
       machineType: N1_HIGHCPU_32
       timeout: 4000s
+- name: debug
+  activation:
+    - command: debug
+  patches:
+    - op: replace
+      # cartservice requires a debug dockerfile
```
kelsk, line 75: How about a more detailed comment that mimics the format of the comment for the "gcb" profile, i.e.: 
```
# "debug" profile replaces the default Dockerfile in cartservice with Dockerfile.debug, 
# which enables debugging via skaffold.

# This profile is used by default when running skaffold debug.
```

j-windsor: Done!

askmeegs (approved):
Tested locally with `skaffold dev` (default profile) and `skaffold debug` (debug profile). LGTM.

PR #610, opened 2021-10-14
Use context.WithTimeout instead of grpc.WithTimeout

jxlwqq:
WithTimeout returns a DialOption that configures a timeout for dialing a ClientConn initially. 

This is valid if and only if WithBlock() is present.

Deprecated: use DialContext and context.WithTimeout instead.


jxlwqq:
@googlebot I signed it!

PR #618, opened 2021-11-09
Memorystore (redis) 6_X instead of 5

mathieu-benoit:
I just tested Memorystore (redis) 6_X with the `cartservice` and it's working fine, so thought we could update our doc with that.

https://cloud.google.com/memorystore/docs/redis/supported-versions


askmeegs (approved):

askmeegs:
Thanks Mathieu! :D

askmeegs (approved):

PR #629, opened 2021-11-20
cartservice - .NET 6

mathieu-benoit:
Update `cartservice` to the new LTS .NET 6.0

https://devblogs.microsoft.com/dotnet/announcing-net-6/


mathieu-benoit:
Tested and deployed http://34.75.58.116, ready for your review, thanks.

src/cartservice/src/cartservice.csproj
```diff
@@ -1,13 +1,13 @@
...
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Grpc.AspNetCore" Version="2.40.0" />
-    <PackageReference Include="Grpc.HealthCheck" Version="2.41.1" />
-    <PackageReference Include="StackExchange.Redis" Version="2.2.79" />
+    <PackageReference Include="Grpc.AspNetCore" Version="2.41.0-pre1" />
```
Shabirmean, line 8: @mathieu-benoit Any reason we want to update to the `pre release` version? Is there a specific feature you are hoping to use?

Shabirmean (commented):
Thank you for the changes. I've left a small comment related to adding the pre-release version

src/cartservice/src/cartservice.csproj
mathieu-benoit: Thanks for asking @Shabirmean, I should have been more prescriptive on describing why I was using this pre version.
Actually there is an actual issue with the .NET GRPC: https://github.com/grpc/grpc-dotnet/issues/1470 which for now just fixed with this https://github.com/grpc/grpc-dotnet/releases/tag/v2.41.0-pre1.
I think it's safe to move forward with that (and I could do another PR later to update it accordingly), or we could wait a more official version before merging this PR in master/main branch.
Let me know what you prefer.

Shabirmean: Thank you @mathieu-benoit for sharing that extra context! I think this is good to be merged. 

I have opened the issue #632 to ensure we keep track of this change! If you happen to update it in a follow up PR you can refer to it!

Shabirmean (approved):
LGTM!

Issue #538, opened 2021-05-03
Possible memory leak in NodeJS / Python services


askmeegs:
Uptime checks for the production deployment of OnlineBoutique have been failing once every few weeks. Looking at `kubectl events` timed with an uptime check failure -- 

```
38m         Warning   NodeSysctlChange   node/gke-online-boutique-mast-default-pool-65a22575-azeq   {"unmanaged": {"net.ipv4.tcp_fastopen_key": "004baa97-3c3b554d-9bbcccf8-870ced36"}}
43m         Warning   NodeSysctlChange   node/gke-online-boutique-mast-default-pool-65a22575-i6m8   {"unmanaged": {"net.ipv4.tcp_fastopen_key": "706b7d5f-9df4b412-e8eb875e-179c4765"}}
46m         Warning   NodeSysctlChange   node/gke-online-boutique-mast-default-pool-65a22575-jvwz   {"unmanaged": {"net.ipv4.tcp_fastopen_key": "a0f734c5-5c9a56e1-06aeb420-0010498e"}}
39m         Warning   OOMKilling         node/gke-online-boutique-mast-default-pool-65a22575-jvwz   Memory cgroup out of memory: Kill process 569290 (node) score 2181 or sacrifice child
Killed process 569290 (node) total-vm:1418236kB, anon-rss:121284kB, file-rss:33236kB, shmem-rss:0kB
39m         Warning   OOMKilling         node/gke-online-boutique-mast-default-pool-65a22575-jvwz   Memory cgroup out of memory: Kill process 2592522 (grpc_health_pro) score 1029 or sacrifice child
Killed process 2592530 (grpc_health_pro) total-vm:710956kB, anon-rss:1348kB, file-rss:7376kB, shmem-rss:0kB
```

It looks like memory requests are exceeding their limit. There seems to be plenty of allocatable memory across the prod GKE nodes 
<img width="1400" alt="Screen Shot 2021-05-03 at 1 55 56 PM" src="https://user-images.githubusercontent.com/3137106/116916179-f9806300-ac1a-11eb-84e6-66864d8d559f.png">

But as observed by @bourgeoisor, it seems that three of the workloads are using steadily increasing amounts of memory until the pods are killed by GKE. 

Currency and payment (NodeJS):

[...]

PR #654, opened 2021-12-09
test: has no debug (do not merge)

Shabirmean:
### Background 
- Testing memory-leak related to #538


Shabirmean:
Heap usage over two days for currency service without `debug`

image

image

PR #666, opened 2021-12-17
Releasing v0.3.4 to include log4j 2.16.0

donmccasland:
### Background 
New Log4j exploits were found in version 2.15.0.  We updated our code to use 2.16.0.  Now we need to create a release.


askmeegs:
Holding this until we actually upgrade log4j to 2.16 in adservice : )

Issue #691, opened 2022-01-19
Fix shipping cost calculation formula


Pawlaklis:
### Describe the bug 
Right now the cost of shipping is calculated using formula
<img src="https://latex.codecogs.com/svg.image?price&space;=&space;n&plus;3^{1&plus;\frac{n}{5}}" title="price = n+3^{1+\frac{n}{5}}" />

Because of that the costs grows exponential with number of products which is not only very unusual way to calculate shipping but also creates absurdly large numbers.

### To Reproduce 
Example: When buying 50 T-shirts (which costs $949.50) the shipping price is $177196.99 which is 190 times more that price of products.

### Screenshots 
![Screenshot 2022-01-17 at 19-32-24 Online Boutique](https://user-images.githubusercontent.com/55978132/150124219-6b2e1ea1-4721-4249-abbc-df9a9a843039.png)

### Possible solution
Calculate the cost as flat number not dependant on quantity of bought products.

NimJay:
Thanks for point this out, @Pawlaklis!
@ckim328 has gone ahead and created a solution in #694.
We're just using a fixed shipping cost now to _keep things simple_.

PR #694, opened 2022-01-20
Shipping fee to be a flat rate

ckim328:
### Background 
Related to https://github.com/GoogleCloudPlatform/microservices-demo/issues/691 , where it was found that the shipping cost was calculated using  As per the suggestion in the issue, I changed the shipping fee to be a 'flat rate'

### Fixes 
Related to this issue https://github.com/GoogleCloudPlatform/microservices-demo/issues/691
### Change Summary
Changed `quoteByCountFloat` in the shippingService to return `8.99` (an arbitrary number, but shipping usually costs around this much in my experience)

### Additional Notes
If we did want to calculate the shipping rate still, I'm ok with closing this PR!

### Testing Procedure
Tested it locally, and shows the shipping rate to be `8.99`

### DIFF
--- a/src/shippingservice/shippingservice_test.go
+++ b/src/shippingservice/shippingservice_test.go
@@ -50,7 +50,7 @@ func TestGetQuote(t *testing.T) {
 	if err != nil {
 		t.Errorf("TestGetQuote (%v) failed", err)
 	}
-	if res.CostUsd.GetUnits() != 11 || res.CostUsd.GetNanos() != 220000000 {
+	if res.CostUsd.GetUnits() != 8 || res.CostUsd.GetNanos() != 990000000 {
 		t.Errorf("TestGetQuote: Quote value '%d.%d' does not match expected '%s'", res.CostUsd.GetUnits(), res.CostUsd.GetNanos(), "11.220000000")
 	}
 }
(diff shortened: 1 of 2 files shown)


src/shippingservice/shippingservice_test.go
```diff
@@ -50,7 +50,7 @@ func TestGetQuote(t *testing.T) {
 	if err != nil {
 		t.Errorf("TestGetQuote (%v) failed", err)
 	}
-	if res.CostUsd.GetUnits() != 11 || res.CostUsd.GetNanos() != 220000000 {
+	if res.CostUsd.GetUnits() != 8 || res.CostUsd.GetNanos() != 990000000 {
 		t.Errorf("TestGetQuote: Quote value '%d.%d' does not match expected '%s'", res.CostUsd.GetUnits(), res.CostUsd.GetNanos(), "11.220000000")
```
bourgeoisor, line 54: Should the expected be changed to 8.99 too?

ckim328: i changed it in my second commit!

NimJay (approved):
I love this solution.
It's simple — as demo apps should be. :)

Changes look good to me.
I've tested the staging URL with different currencies. 👍

PR #698, opened 2022-01-28
Complete PR#689 - Cloud Armor stup for onlineboutique.dev

mathieu-benoit:
- Fix the `defaultBackend` field of the Ingress to avoid any old pointer to the old `frontend-nodeport` service

### DIFF
--- a/.github/release-cluster/frontend-ingress.yaml
+++ b/.github/release-cluster/frontend-ingress.yaml
@@ -7,6 +7,11 @@ metadata:
     networking.gke.io/managed-certificates: online-boutique-certificate
     networking.gke.io/v1beta1.FrontendConfig: frontend-frontend-config
 spec:
+  defaultBackend:
+    service:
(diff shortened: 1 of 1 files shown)


NimJay (approved):
Previous thread: https://github.com/GoogleCloudPlatform/microservices-demo/pull/689#issuecomment-1024391665

I just checked `kubectl get service frontend-nodeport` and got:
```
Error from server (NotFound): services "frontend-nodeport" not found
```
which is what I expect. No more `frontend-nodeport`. :)
I also see that the [Cloud Armor policy has a target](https://pantheon.corp.google.com/net-security/securitypolicies/details/online-boutique-security-policy?project=online-boutique-ci&organizationId=655675970190&tab=rules&networkCloudArmorSecurityPolicyRulesTablesize=50&cloudshell=false).
 
Thanks again for taking care of this, @mathieu-benoit.
Approved.

Issue #697, opened 2022-01-28
Frontend blocking requests when not in GCP


prune998:
### Describe the bug 

a request against the `frontend` app when deployed in a non-GCP environment lead to a 2 minutes delay

### To Reproduce 

- deploy the whole project in a K8s cluster WITHOUT an API answering on `169.254.169.254:80`. This can be achieved using a K8s cluster out of AWS or GCP, or, like me, using Cilium Network Policies to block such traffic
- curl against the `frontend` service... it will block for ~ 2 minutes

### Logs  

```
{"message":"Tracing disabled.","severity":"info","timestamp":"2022-01-28T18:43:52.381216858Z"}
{"message":"Profiling disabled.","severity":"info","timestamp":"2022-01-28T18:43:52.381283518Z"}
{"message":"starting server on :8080","severity":"info","timestamp":"2022-01-28T18:43:52.381944193Z"}
{"http.req.id":"ec621cfe-d9b8-4965-aa94-bfd8033c15c5","http.req.method":"GET","http.req.path":"/","message":"request started","session":"d5cf17c3-c869-4898-bb58-efa12d1fb2dd","severity":"debug","timestamp":"2022-01-28T18:44:18.181114709Z"}
{"currency":"USD","http.req.id":"ec621cfe-d9b8-4965-aa94-bfd8033c15c5","http.req.method":"GET","http.req.path":"/","message":"home","session":"d5cf17c3-c869-4898-bb58-efa12d1fb2dd","severity":"info","timestamp":"2022-01-28T18:44:18.181157082Z"}
{"http.req.id":"ec621cfe-d9b8-4965-aa94-bfd8033c15c5","http.req.method":"GET","http.req.path":"/","message":"ENV_PLATFORM is: aws","session":"d5cf17c3-c869-4898-bb58-efa12d1fb2dd","severity":"debug","timestamp":"2022-01-28T18:44:18.346517667Z"}
{"http.req.id":"ec621cfe-d9b8-4965-aa94-bfd8033c15c5","http.req.method":"GET","http.req.path":"/","message":"Failed to fetch the name of the cluster in which the pod is runningcompute: Received 503 `upstream connect error or disconnect/reset before headers. reset reason: connection failure`","session":"d5cf17c3-c869-4898-bb58-efa12d1fb2dd","severity":"error","timestamp":"2022-01-28T18:45:20.438611873Z"}
[...]

Issue #685, opened 2022-01-11
Timeout when not running on GCE


CapacitorSet:
### Describe the bug 
When the application is deployed on [Kind](https://kind.sigs.k8s.io/) (and, I believe, any non-GCE platform), requests time out.

I think the root cause is that the frontend calls `getDeploymentDetails` on every request, which uses `cloud.google.com/go/compute/metadata` and makes a request to `http://169.254.169.254/computeMetadata/...`. This IP however does not exist on other platforms, causing delays of several minutes in HTTP requests.

This appears to be a regression from 09956ca4.

### To Reproduce 
1. Install Kind
2. Create a Kind node: `kind create cluster --kubeconfig=/tmp/config --name=cluster1`
3. In the Kind node, apply `release/kubernetes-manifests.yaml` from v0.3.5
4. Make a request to the frontend
5. Wait for the request to terminate; if no timeout is applied, it will return after 10 minutes with the homepage.

### Logs  
Frontend logs:
```
{
  "http.req.id": "8fccb933-80d9-4482-badf-a3963f35a13f",
  "http.req.method": "GET",
  "http.req.path": "/",
  "message": "Failed to fetch the name of the cluster in which the pod is runningGet \"http://169.254.169.254/computeMetadata/v1/instance/attributes/cluster-name\": dial tcp 169.254.169.254:80: i/o timeout",
  "session": "0706fb66-c1aa-4215-a93a-8db82f586e7c",
  "severity": "error",
  "timestamp": "2022-01-11T14:44:21.017021115Z"
}
{
  "http.req.id": "8fccb933-80d9-4482-badf-a3963f35a13f",
  "http.req.method": "GET",
  "http.req.path": "/",
  "message": "Failed to fetch the Zone of the node where the pod is scheduledGet \"http://169.254.169.254/computeMetadata/v1/instance/zone\": dial tcp 169.254.169.254:80: i/o timeout",
  "session": "0706fb66-c1aa-4215-a93a-8db82f586e7c",
  "severity": "error",
  "timestamp": "2022-01-11T14:47:22.011728459Z"
}
```

### Environment 
 - OS: Arch Linux
 - Kubernetes distribution, version: Kind v0.11.1

### Exposure 
This bug occurs on every deployment. I believe it to be the reason for end-to-end tests in our CI failing due to timeouts.

saamalik:
I can confirm this behavior. I reverted to release `v0.3.1` to get around this. Thx!

[...]

Issue #628, opened 2021-11-19
Indicate Cloud Region/Location in Front-end


alizaidis:
Adding tutorial to showcase this application running in a multi-cluster, multi-region setup. Need an easy way for the user to tell the location/region of the pod responding to the client request.

Working on a fix for this. PR incoming shortly.

askmeegs:
Thanks Ali! You may be able to port what we have for this in the Bank of Anthos app. 
https://github.com/GoogleCloudPlatform/bank-of-anthos/blob/master/src/frontend/frontend.py#L68

askmeegs:
Note to whoever takes this -- let's do cluster, region, and zone for good measure.
If not GKE, list "unknown" for now.

Shabirmean:
@alizaidis  This is fixes in https://github.com/GoogleCloudPlatform/microservices-demo/pull/636

PR #699, opened 2022-01-28
Fix Deployment Details Bug

NimJay:
* This fixes: #697 and #685.

## Background

* Look at the footer of [onlineboutique.dev](https://onlineboutique.dev/).
* You'll see details about the "Zone", "Pod", and "Cluster".
* This info is loaded by the `getDeploymentDetails` (upon every HTTP request) in the `frontend` service. 
* `getDeploymentDetails` is using a Google Cloud API (cloud.google.com/go/compute/metadata) to load the cluster name and zone.
* This was causing issues on non-GCP deployments (e.g., [Kind](https://kind.sigs.k8s.io/) and EKS). 
* I've fixed the issue by doing 2 things: 
  * perform `getDeploymentDetails` asynchronously (so that the HTTP request handler isn't blocked)
  * have `getDeploymentDetails` load necessary info _once_.
* This certainly isn't a perfect solution — but fixing this issue is **high** priority. 

## Summary of Changes

**1. New File: deployment_details.go**

* I've moved the logic related to loading deployment details into a new file called `deployment_details.go`.
* Reason 1: Single responsibility principle. 
* Reason 2: The `handlers.go` file is too long in my opinion.

**2. Don't Wait on Deployment Details**
* Previously, we loaded deployment details synchronously.
* But now, we're using a `go` routine to load deployment details asynchronously.

**3. Only Load Deployment Details Once**
* Previously, every HTTP request that needed to display deployment details reloaded the deployment details (e.g., each request made a call to Google Cloud to figure out the zone and cluster name).
* Now we're only loading it once!

**4. Mutex**
* I have introduced the use of a mutex to ensure that two HTTP requests don't race to load the deployment details twice.
* Basically, I want to avoid this scenario:
  * Request 1: Are the details loaded? Nope, gotta load it.
  * Request 2: Are the details loaded? Nope, gotta load it.
  * Request 1: Let everyone know I'm taking care of loading it.
  * Request 2: Let everyone know I'm taking care of loading it.

**5. Front-end Message**
* The following message is now displayed in the footer appropriately:
> Deployment details are either still loading or failed to load.
> Try refreshing this page.

## Testing
* I have tested these changes on both a GKE cluster (e.g., see staging URL) and a local Kind cluster (i.e., see comment below). This works. 👍 
* If you would like to test these changes yourself on a Kind cluster:
1. Install Kind: `brew install kind`
1. Create a Kind cluster: `kind create cluster`
1. Make sure you're using the Kind cluster: `kubectx`. You should see `kind-kind` highlighted. 
1. Pull the changes from this pull-request: `git pull nimjay/non-gcp-bug`.
1. Build the modified frontend service: `docker push `
1. Push the modified frontend service: `docker push gcr.io/my-project/new-front-end`
1. Change the `kubernetes-manifests.yaml` file: use `docker push gcr.io/my-project/new-front-end` instead of `gcr.io/google-samples/microservices-demo/frontend:v0.3.5`.
1. `kubectl apply -f kubernetes-manifests.yaml`
1. To access the deployment on your web browser as `localhost:8080`, run: `kubectl port-forward <name-of-frontend-pod> 8080:8080`. (Use `kubectl get pods` to get the name of your front-end pod.)

## Additional Notes
* I am very new to Golang — so I'm open to significantly refactoring my code. I'm guessing there's a TON of room for improvement. 😅

### DIFF
--- a/src/frontend/deployment_details.go
+++ b/src/frontend/deployment_details.go
@@ -15,6 +15,8 @@ var log *logrus.Logger
 
 func init() {
 	initializeLogger()
+	// Use a goroutine to ensure loadDeploymentDetails()'s GCP API
+	// calls don't block non-GCP deployments. See issue #685.
 	go loadDeploymentDetails()
 }
 
--- a/src/frontend/deployment_details.go
+++ b/src/frontend/deployment_details.go
@@ -9,8 +9,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var areDeploymentDetailsLoaded = false
-var deploymentDetailsMap = make(map[string]string)
+var deploymentDetailsMap map[string]string
 var log *logrus.Logger
 
 func init() {
@@ -34,7 +33,8 @@ func initializeLogger() {
 	log.Out = os.Stdout
 }
 
-func loadDeploymentDetails() map[string]string {
+func loadDeploymentDetails() {
+	deploymentDetailsMap = make(map[string]string)
 	var metaServerClient = metadata.NewClient(&http.Client{})
 
 	podHostname, err := os.Hostname()
@@ -61,15 +61,8 @@ func loadDeploymentDetails() map[string]string {
 		"zone":     podZone,
 		"hostname": podHostname,
 	}).Debug("Loaded deployment details")
-
-	areDeploymentDetailsLoaded = true
-
-	return deploymentDetailsMap
 }
 
-func getDeploymentDetailsIfLoaded(httpRequest *http.Request) map[string]string {
-	if areDeploymentDetailsLoaded {
-		return deploymentDetailsMap
-	}
-	return nil
+func getDeploymentDetailsIfLoaded() map[string]string {
+	return deploymentDetailsMap
 }
--- a/src/frontend/deployment_details.go
+++ b/src/frontend/deployment_details.go
@@ -3,27 +3,37 @@ package main
 import (
 	"net/http"
 	"os"
-	"sync"
+	"time"
 
 	"cloud.google.com/go/compute/metadata"
 	"github.com/sirupsen/logrus"
 )
 
-const (
-	DetailsAreNotLoaded = 0
-	DetailsAreLoading   = 1
-	DetailsAreLoaded    = 2
-)
-
+var areDeploymentDetailsLoaded = false
 var deploymentDetailsMap = make(map[string]string)
-var detailsLoadingStatus = DetailsAreNotLoaded
-var detailsLoadingStatusMutex sync.Mutex // Ensures only 1 HTTP request loads deployment details.
+var log *logrus.Logger
 
-func loadDeploymentDetails(httpRequest *http.Request) map[string]string {
+func init() {
+	initializeLogger()
+	go loadDeploymentDetails()
+}
+
+func initializeLogger() {
+	log = logrus.New()
+	log.Level = logrus.DebugLevel
+	log.Formatter = &logrus.JSONFormatter{
+		FieldMap: logrus.FieldMap{
+			logrus.FieldKeyTime:  "timestamp",
+			logrus.FieldKeyLevel: "severity",
+			logrus.FieldKeyMsg:   "message",
+		},
+		TimestampFormat: time.RFC3339Nano,
+	}
+	log.Out = os.Stdout
+}
+
+func loadDeploymentDetails() map[string]string {
 	var metaServerClient = metadata.NewClient(&http.Client{})
-
-	// The use of httpRequest here lets us to see HTTP request info in the logs.
-	var log = httpRequest.Context().Value(ctxKeyLog{}).(logrus.FieldLogger)
 
 	podHostname, err := os.Hostname()
 	if err != nil {
@@ -50,18 +60,14 @@ func loadDeploymentDetails(httpRequest *http.Request) map[string]string {
 		"hostname": podHostname,
 	}).Debug("Loaded deployment details")
 
-	detailsLoadingStatus = DetailsAreLoaded
+	areDeploymentDetailsLoaded = true
 
 	return deploymentDetailsMap
 }
 
 func getDeploymentDetailsIfLoaded(httpRequest *http.Request) map[string]string {
-	detailsLoadingStatusMutex.Lock()
-	defer detailsLoadingStatusMutex.Unlock()
-	if detailsLoadingStatus == DetailsAreNotLoaded {
-		detailsLoadingStatus = DetailsAreLoading
-		go loadDeploymentDetails(httpRequest)
-		return nil
+	if areDeploymentDetailsLoaded {
+		return deploymentDetailsMap
 	}
-	return deploymentDetailsMap
+	return nil
 }
--- a/src/frontend/deployment_details.go
+++ b/src/frontend/deployment_details.go
@@ -1,0 +1,67 @@
+package main
+
+import (
+	"net/http"
+	"os"
+	"sync"
+
+	"cloud.google.com/go/compute/metadata"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	DetailsAreNotLoaded = 0
+	DetailsAreLoading   = 1
+	DetailsAreLoaded    = 2
+)
+
+var deploymentDetailsMap = make(map[string]string)
+var detailsLoadingStatus = DetailsAreNotLoaded
+var detailsLoadingStatusMutex sync.Mutex // Ensures only 1 HTTP request loads deployment details.
+
+func loadDeploymentDetails(httpRequest *http.Request) map[string]string {
+	var metaServerClient = metadata.NewClient(&http.Client{})
+
+	// The use of httpRequest here lets us to see HTTP request info in the logs.
+	var log = httpRequest.Context().Value(ctxKeyLog{}).(logrus.FieldLogger)
+
+	podHostname, err := os.Hostname()
+	if err != nil {
+		log.Error("Failed to fetch the hostname for the Pod", err)
+	}
+
+	podCluster, err := metaServerClient.InstanceAttributeValue("cluster-name")
+	if err != nil {
+		log.Error("Failed to fetch the name of the cluster in which the pod is running", err)
+	}
+
+	podZone, err := metaServerClient.Zone()
+	if err != nil {
+		log.Error("Failed to fetch the Zone of the node where the pod is scheduled", err)
+	}
+
+	deploymentDetailsMap["HOSTNAME"] = podHostname
+	deploymentDetailsMap["CLUSTERNAME"] = podCluster
+	deploymentDetailsMap["ZONE"] = podZone
+
+	log.WithFields(logrus.Fields{
+		"cluster":  podCluster,
+		"zone":     podZone,
+		"hostname": podHostname,
+	}).Debug("Loaded deployment details")
+
+	detailsLoadingStatus = DetailsAreLoaded
+
+	return deploymentDetailsMap
+}
+
+func getDeploymentDetailsIfLoaded(httpRequest *http.Request) map[string]string {
+	detailsLoadingStatusMutex.Lock()
+	defer detailsLoadingStatusMutex.Unlock()
+	if detailsLoadingStatus == DetailsAreNotLoaded {
+		detailsLoadingStatus = DetailsAreLoading
+		go loadDeploymentDetails(httpRequest)
+		return nil
+	}
+	return deploymentDetailsMap
+}
--- a/src/frontend/deployment_details.go
+++ b/src/frontend/deployment_details.go
@@ -62,7 +62,3 @@ func loadDeploymentDetails() {
 		"hostname": podHostname,
 	}).Debug("Loaded deployment details")
 }
-
-func getDeploymentDetailsIfLoaded() map[string]string {
-	return deploymentDetailsMap
-}
--- a/src/frontend/handlers.go
+++ b/src/frontend/handlers.go
@@ -115,7 +115,7 @@ func (fe *frontendServer) homeHandler(w http.ResponseWriter, r *http.Request) {
 		"platform_css":      plat.css,
 		"platform_name":     plat.provider,
 		"is_cymbal_brand":   isCymbalBrand,
-		"deploymentDetails": getDeploymentDetailsIfLoaded(r),
+		"deploymentDetails": getDeploymentDetailsIfLoaded(),
 	}); err != nil {
 		log.Error(err)
 	}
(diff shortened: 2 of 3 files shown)


src/frontend/deployment_details.go
```diff
@@ -0,0 +1,67 @@
...
+
+var deploymentDetailsMap = make(map[string]string)
+var detailsLoadingStatus = DetailsAreNotLoaded
+var detailsLoadingStatusMutex sync.Mutex // Ensures only 1 HTTP request loads deployment details.
+
+func loadDeploymentDetails(httpRequest *http.Request) map[string]string {
+	var metaServerClient = metadata.NewClient(&http.Client{})
+
+	// The use of httpRequest here lets us to see HTTP request info in the logs.
+	var log = httpRequest.Context().Value(ctxKeyLog{}).(logrus.FieldLogger)
```
NimJay, line 26: I think, ideally, this `loadDeploymentDetails` function would not know about the `httpRequest` object.
So, in the future, we could pass in the `log` object instead of the `httpRequest` object into the function.
But such an improvement would be outside the scope of this pull-request.

NimJay:
I have tested this on a Kind cluster (i.e., a non-GCP cluster) and it works! 😌
Ping me if you would like me to quickly showcase this:

Screen Shot 2022-01-31 at 10 50 03 AM

The footer displayed the following message: 
> Deployment details are either still loading or failed to load. Try refreshing this page.
I think we should eventually work actually displaying deployment details for non-GCP clusters (instead of the above message).
But this is out of scope for now — let's fix this P1 issue.

ckim328:
I just tried with minikube, and footer had `Pod` and `Cluster` shown as blank (instead of _Deployment details are either still loading or failed to load. Try refreshing this page._) Is this the expected behavior?
Screen Shot 2022-01-31 at 11 53 38 AM

src/frontend/deployment_details.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"cloud.google.com/go/compute/metadata"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	DetailsAreNotLoaded = 0
+	DetailsAreLoading   = 1
+	DetailsAreLoaded    = 2
+)
```
NimJay, line 16: I was inspired to do enum this way by [this article, Using Enums (and Enum Types) in Golang
](https://www.sohamkamani.com/golang/enums/).

NimJay:
@ckim328, thanks for testing these changes on `minikube`.
Ah, yes, I'm seeing the same output on Kind.
Thanks for catching that!

Screen Shot 2022-01-31 at 12 47 49 PM

**What's happening?**
The following message is only displayed while `getDeploymentDetails()` is still loading (i.e., still trying to talk to Google Cloud to figure out zone/cluster info).
> "Deployment details are either still loading or failed to load. Try refreshing this page."
The changes from this pull-request ensures that the `frontend` service only tries to talk to Google Cloud once:
So _after_ talking to Google Cloud fails, we will see the following in the `kubectl logs` of the frontend pod:
> ... Failed to fetch the name of the cluster in which the pod is running ...
> ... Failed to fetch the Zone of the node where ...

So yes, this is expected. 

If you want, instead of showing an empty value for zone and cluster, I can state, "failed to load" (or a similar message) or I can hide the "zone" and "cluster" labels completely?
What do you think?

ckim328:
Thanks for the clarification @NimJay 👍  
I think the case where it is a locally deployed, I'm ok with removing the labels

 It's ok if it's not captured in the scope of this P1 discussion!

src/frontend/templates/footer.html
```diff
@@ -32,6 +32,9 @@
                     <b>Cluster: </b>{{index .deploymentDetails "CLUSTERNAME" }}<br/>
                     <b>Zone: </b>{{index .deploymentDetails "ZONE" }}<br/>
                     <b>Pod: </b>{{index .deploymentDetails "HOSTNAME" }}
+                    {{ else }}
+                    Deployment details are either still loading or failed to load.
```
bourgeoisor, line 36: Not too opiniated on this but I think we should silently fail and not show anything rather than show an error message. Or just show "Deployment details are unavailable" or similar.
The reasoning behind this is that more often than not, the deployment details are simply not available, but the current diff makes it seem like the user _should_ be reloading, or that there is something fundamentally wrong with their cluster's well being.
Thoughts?

src/frontend/deployment_details.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	podZone, err := metaServerClient.Zone()
+	if err != nil {
+		log.Error("Failed to fetch the Zone of the node where the pod is scheduled", err)
+	}
+
+	deploymentDetailsMap["HOSTNAME"] = podHostname
+	deploymentDetailsMap["CLUSTERNAME"] = podCluster
+	deploymentDetailsMap["ZONE"] = podZone
```
bourgeoisor, line 45: Why don't we just get this information once, and only once ever? We can do this from an `init()` function, with a package variable that we can query as many times as we want.

```
// deployment_details.go
package main

import (...)

var deploymentDetails map[string]string

func init() {
    // get the deployment details from the meta server

    deploymentDetails = map[string]string {
        "HOSTNAME": podHostname,
        "CLUSTERNAME": podCluster,
        "ZONE": podZone,
    }
}
```

Then from the handlers:
```
if err := templates.ExecuteTemplate(w, "home", map[string]interface{}{
    . . .
    . . .
    "deploymentDetails": deploymentDetails,
}
```

That would mean no mutex, no enums, no race conditions, no "loading / not loading", and reduce complexity.

bourgeoisor (changes requested):

src/frontend/deployment_details.go
bourgeoisor: If the meta server is not available or reachable (e.g. if you're deploying locally), then you would leave the map blank, and that should be taken care of seamlessly by the HTML template with the "if deploymentDetails" bit.

src/frontend/deployment_details.go
```diff
@@ -0,0 +1,73 @@
...
+	log.Formatter = &logrus.JSONFormatter{
+		FieldMap: logrus.FieldMap{
+			logrus.FieldKeyTime:  "timestamp",
+			logrus.FieldKeyLevel: "severity",
+			logrus.FieldKeyMsg:   "message",
+		},
+		TimestampFormat: time.RFC3339Nano,
+	}
+	log.Out = os.Stdout
+}
```
NimJay, line 33: I copied this log-related chunk from https://github.com/GoogleCloudPlatform/microservices-demo/blob/main/src/shippingservice/main.go#L46.

src/frontend/deployment_details.go
NimJay: Great suggestion! I'm glad you took a look at my changes.
I'm now loading the deployment details from `func init()`.
I was able to get rid of the following complexities: 
* mutex
* race conditions
* enums

However, I wasn't able to get rid of the "no 'loading / not loading'" stuff. This is because we don't want the loading of details to block the main thread in non-GCP deployments (as described in #685).

src/frontend/templates/footer.html
NimJay: @ckim328 suggested a similar "fail silently" user experience (i.e., hiding labels that have no values).
So I've gone ahead and implemented that. :)

I have removed the following verbiage: 
>  or failed to load

because this was not true. If an error occurred, we would've just seen labels with no values.

I have preserved the following verbiage:
> Deployment details are still loading.
> Try refreshing this page.

because good error message are actionable.
And there is a tutorial that asks developers to look at the deployment details (see #628).

Thanks, you two.

src/frontend/deployment_details.go
```diff
@@ -3,27 +3,37 @@ package main
...
 )
 
-const (
-	DetailsAreNotLoaded = 0
-	DetailsAreLoading   = 1
-	DetailsAreLoaded    = 2
-)
-
+var areDeploymentDetailsLoaded = false
 var deploymentDetailsMap = make(map[string]string)
```
bourgeoisor, line 13: This can stay as `var deploymentDetailsMap = map[string]string` and then you have an easy way to check if the deployment are loaded / present, which is that `deploymentDetailsMap != nil`.
That would also remove the need for that additional `areDeploymentDetailsLoaded` variable.

src/frontend/deployment_details.go
```diff
@@ -3,27 +3,37 @@ package main
...
-func loadDeploymentDetails(httpRequest *http.Request) map[string]string {
-	var metaServerClient = metadata.NewClient(&http.Client{})
+func init() {
+	initializeLogger()
+	go loadDeploymentDetails()
+}
 
-	// The use of httpRequest here lets us to see HTTP request info in the logs.
-	var log = httpRequest.Context().Value(ctxKeyLog{}).(logrus.FieldLogger)
+func initializeLogger() {
```
bourgeoisor, line 23: I'm okay with having a single logger, but if so:
- Let's put it in either `main.go` or its own `logger.go`, and
- Let's use it everywhere. I notice some files (like `main.go` uses their own duplicated loggers)

src/frontend/deployment_details.go
```diff
@@ -50,18 +60,14 @@ func loadDeploymentDetails(httpRequest *http.Request) map[string]string {
 		"hostname": podHostname,
 	}).Debug("Loaded deployment details")
 
-	detailsLoadingStatus = DetailsAreLoaded
+	areDeploymentDetailsLoaded = true
 
 	return deploymentDetailsMap
 }
 
 func getDeploymentDetailsIfLoaded(httpRequest *http.Request) map[string]string {
```
bourgeoisor, line 70: This is completely redundant if we change the `deploymentDetailsMap` to be `nil` at initialization (so remove the `make()` from there, so this entire method can go away.

src/frontend/deployment_details.go
```diff
@@ -50,18 +60,14 @@ func loadDeploymentDetails(httpRequest *http.Request) map[string]string {
 		"hostname": podHostname,
 	}).Debug("Loaded deployment details")
 
-	detailsLoadingStatus = DetailsAreLoaded
+	areDeploymentDetailsLoaded = true
 
 	return deploymentDetailsMap
```
bourgeoisor, line 67: No need to return anything. The called (`go loadDeploymentDetails()`) is not expecting anything returned. We just have to modify the global map and silently return (which is implicit at the end of the function, so we don't need a `return` at all)

src/frontend/templates/footer.html
```diff
@@ -29,11 +29,17 @@
...
+                        {{ end }}
+                        {{ if index .deploymentDetails "ZONE" }}
+                        <b>Zone: </b>{{ index .deploymentDetails "ZONE" }}<br/>
+                        {{ end }}
+                        {{ if index .deploymentDetails "HOSTNAME" }}
+                        <b>Pod: </b>{{ index .deploymentDetails "HOSTNAME" }}
+                        {{ end }}
                     {{ else }}
-                    Deployment details are either still loading or failed to load.
+                    Deployment details are still loading.
```
bourgeoisor, line 42: What if we're running on e.g. Minikube, where `deploymentDetails` is always going to be `nil`? Should we show "deployment details are still loading" if we know explicitly that it's _never_ going to appear? Why don't we just remove that section entirely, or replace it with "Deployment details unavailable"?

bourgeoisor (changes requested):

src/frontend/deployment_details.go
Shabirmean: I wonder, if we should still account for temporary unreachability of the `metadata server`. I thought the previous mutex approach took care of all possibilities. The current change would render the details never loaded if the first attempt on init fails. 

Just a thought. We could still see how this works and add it as an improvement if need be.

src/frontend/templates/footer.html
NimJay: Good thought.
On non-GCP clusters (e.g., Kind and minikube), the Pod name will be loaded.
Screen Shot 2022-02-01 at 1 42 09 PM
So let's keep this wording.

src/frontend/deployment_details.go
NimJay: Good catch!
I will work on not returning anything. :)

src/frontend/deployment_details.go
NimJay: Great suggestion! 👌
Now the declaration at the top looks like:
```
var deploymentDetailsMap map[string]string
```
and it's initialized in the `loadDeploymentDetails()` function:
```
func loadDeploymentDetails() {
	deploymentDetailsMap = make(map[string]string)
```
and the `getDeploymentDetailsIfLoaded` function is now:
```
func getDeploymentDetailsIfLoaded() map[string]string {
	return deploymentDetailsMap
}
```

src/frontend/deployment_details.go
NimJay: I was thinking the same thing.
Let's track this in a different issue: https://github.com/GoogleCloudPlatform/microservices-demo/issues/703
I'd say it's outside the scope of this P1.

src/frontend/deployment_details.go
NimJay: I've preserved the `getDeploymentDetailsIfLoaded()` because it implies that the thing being returned might not have been loaded; thus, I assume anyone using/seeing `getDeploymentDetailsIfLoaded()` will cautiously check if it returns `nil`.
Thoughts?

NimJay:
@bourgeoisor thanks for the new round of comments! 
Great suggestions. I've responded to all of them — and applied most.
I've tested on a local Kind cluster, and everything works! 👍

src/frontend/templates/footer.html
bourgeoisor: Oh gotcha, that makes sense!!

src/frontend/deployment_details.go
bourgeoisor: That's a fair point, though, in what scenario would the meta server to be unavailable? I thought maybe if the GCE API (https://status.cloud.google.com/) is down completely, but the meta server is internal, no? The SLA for any arbitrary single-zone cluster is `>= 99.5%`, so I'm imagining the meta server to be much higher, though that is solely intuition.

I think that's a very low probability, but even so mutex seems overkill. We could do periodic retries in the go routine until we get a result, for example (though I think that's still overkill).

I think it's safe to leave it as-is, but I'm also opened to adding retries (taking into account that it may retry forever for non-GCP instances).

src/frontend/handlers.go
```diff
@@ -116,7 +115,7 @@ func (fe *frontendServer) homeHandler(w http.ResponseWriter, r *http.Request) {
 		"platform_css":      plat.css,
 		"platform_name":     plat.provider,
 		"is_cymbal_brand":   isCymbalBrand,
-		"deploymentDetails": getDeploymentDetails(r),
+		"deploymentDetails": getDeploymentDetailsIfLoaded(),
```
bourgeoisor, line 118: Minor nit: if the `getDeploymentDetailsIfLoaded()` function simply returns `deploymentDetailsMap` as-is (as it should), why don't we remove the function and replace this line with `"deploymentDetails": deploymentDetailsMap`?

If the map is filled up, then we will pass it to be rendered, and if the map hasn't been filled up yet, we'll just be passing on `nil` but that's okay since we're catching that case in the `{{ if $.deploymentDetails }}` of the render.

NimJay: Done. Just replaced `getDeploymentDetailsIfLoaded()` with `deploymentDetailsMap`.

bourgeoisor (approved):
LGTM, with the caveats of the open comment threads

PR #723, opened 2022-02-15
Updated to protobuf==3.15.0 in src/emailservices/requirements.in & src/recommendation/requirements.in

ckim328:
### Background 
Related to https://github.com/GoogleCloudPlatform/microservices-demo/pull/717, where a dependabot alert was generated for `protobuf`. There is also an alert generated for `recommendationservice`, which was captured in this PR
### Fixes 
The dependabot didn't update `requirements.in` file. The PR generated was only for the auto-generated `requirements.txt`. If a pip-compile was ran, it will return `protobuf` to `3.13.0` (I checked the libs that it depends on, and the minimum requirement by any of them was `3.13.0`)
### Change Summary
Pinned `protobuf` in emailservice and recommendationservice to 3.15.0

### Additional Notes
n/a 
### Testing Procedure
n/a

### Related PRs or Issues 
https://github.com/GoogleCloudPlatform/microservices-demo/pull/717


ckim328:
I'll capture the recommendationservice bump in this PR as well

bourgeoisor (approved):
re: dependabot not working with pip-compile by default, it seems possible to enable that: https://github.blog/changelog/2021-01-19-dependabot-pip-compile-5-5-0-support/

Changes look good! I played around with the staging deployment and all looks well.

Before we merge, could we add a little comment besides the `protobuf` bump in the `.in` files that we should remove that line once we bump to a version of the Google APIs that makes use of those newer protobuf?

Since `protobuf` is a transitive dep, we ideally wouldn't have it in the `.in` and would keep it as slim as possible (bumping the direct dep which depends on it, instead)

ckim328:
> re: dependabot not working with pip-compile by default, it seems possible to enable that: https://github.blog/changelog/2021-01-19-dependabot-pip-compile-5-5-0-support/
> 
> Changes look good! I played around with the staging deployment and all looks well.
> 
> Before we merge, could we add a little comment besides the `protobuf` bump in the `.in` files that we should remove that line once we bump to a version of the Google APIs that makes use of those newer protobuf?
> 
> Since `protobuf` is a transitive dep, we ideally wouldn't have it in the `.in` and would keep it as slim as possible (bumping the direct dep which depends on it, instead)

Good idea - I'll make the changes and push it 👍

Issue #776, opened 2022-03-04
redis-cart service and deployment not using port names causes problems with Istio


tiswanso:
### Describe the bug 

In the release manifest, the redis-cart k8s service and deployment don't use [Istio's explicit protocol selection compatible port definitions](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/), e.g. name format `tcp-*`.  

When deployed with Istio auto-sidecar injection, the cartservice isn't able to connect to the redis-cart service.

### To Reproduce 

1. In a cluster with Istio auto-sidecar injection enabled for the default namespace, deploy the `./release/kubernetes-manifests.yaml`
2. The redis-cart and cartservice will come up with Istio sidecars (in the Istio servicemesh) and the cartservice is unable to connect to the redis-cart DB.

PR #777, opened 2022-03-04
Use Istio compatible port name for redis-cart service and deployment

tiswanso:
### Background 

Using [explicit Istio port protocol compatible port names](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/) allows it to correctly determine the protocol.  

This fix uses the explicit `tcp-*` format to ensure compatibility with Istio versions prior to recognizing the `redis` port-name.

### Fixes 

Fixes: #776 

### Testing Procedure

Enable Istio with auto-sidecar injection in the namespace where you're installing the microservices-demo manifest.  Install the manifest and validate all pods come up with Istio sidecars.  Validate the cartservice connects to the `redis-cart` DB.

### DIFF
--- a/release/kubernetes-manifests.yaml
+++ b/release/kubernetes-manifests.yaml
@@ -652,6 +652,8 @@ spec:
         image: redis:alpine
         ports:
         - containerPort: 6379
+          name: tcp-redis
+          protocol: TCP
         readinessProbe:
           periodSeconds: 5
           tcpSocket:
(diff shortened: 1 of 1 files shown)


ckim328:
thank you for the contribution @tiswanso !
Looping in @NimJay for visibility 👀 
We have an automated way for generating the release manifests, so we may be in favour of closing this PR. Thoughts @NimJay ?

PR #799, opened 2022-04-07
Move YAML document separator to top of release manifests

dsbibby:
### Background 
I propose moving the YAML document separator inserted as part of `make-release-artifacts.sh` to the top of each manifest rather than the bottom.

Having the separator at the end of the manifest results in the generated release YAML files containing an empty document at the end. Whilst this is valid YAML, it is untidy, and certain YAML tools (`yq`) have been seen to handle empty document elements badly.

### Change Summary
Moved the YAML separator `---` to before each service manifest rather than after

### Testing Procedure
Tested and deployed generated manifests succesfully

### DIFF
--- a/hack/make-release-artifacts.sh
+++ b/hack/make-release-artifacts.sh
@@ -51,12 +51,12 @@ read_manifests() {
     dir="$1"
 
     while IFS= read -d $'\0' -r file; do
+        echo "---"
+        
         # strip license headers (pattern "^# ")
         awk '
         /^[^# ]/ { found = 1 }
         found { print }' "${file}"
-
-        echo "---"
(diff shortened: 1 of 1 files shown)


ckim328 (approved):

NimJay:
Sounds good to me!
I have not manually tested these changes.
But if anything breaks, we will find when we create the next release (via [these steps](https://github.com/GoogleCloudPlatform/microservices-demo/tree/main/hack#hack)) — which I will likely do in the coming weeks. Any breakage won't impact any users (or other stakeholders) of this repo.
Merging.

Thanks, @ckim328!
And thanks, @dsbibby!

NimJay (approved):

PR #809, opened 2022-04-21
Set up Major.Minor version only for unit tests to avoid too many main…

mathieu-benoit:
Set up Major.Minor version only for unit tests to avoid too many maintenance with release/patch.

Go and .NET unit tests only.


NimJay (approved):
All checks are passing.
And not specifying the patch here sounds good to me.

Issue #650, opened 2021-12-07
misspelled RecommendationService in architecture


fty4:
### Issue
The `RecommendationService` is misspelled in the architecture image.

Instead of writing _recommendation_ with two `m`´s its only one.

### To Reproduce 
1. open https://github.com/GoogleCloudPlatform/microservices-demo/blob/f703536128463e1327bd67f30e9bdb7160e7931b/docs/img/architecture-diagram.png

### Screenshots 
<img width="350px" src="https://user-images.githubusercontent.com/16336640/145055151-86e1fc01-18fd-4137-bafa-fbb9f0ebe72d.png"/>

### Additional context 
Unfortunately there is no SVG provided to create a PR by myself.

mathieu-benoit:
Thanks for raising this, it's now fixed with the new architecture diagram: https://github.com/GoogleCloudPlatform/microservices-demo/blob/main/docs/img/architecture-diagram.png

PR #829, opened 2022-05-13
Update architecture-diagram.png

NimJay:
### Change Summary
* Update the diagram at `docs/img/architecture-diagram.png`.
* See [new diagram here](https://github.com/GoogleCloudPlatform/microservices-demo/tree/nimjay-arch-diagram#architecture).
* See [old diagram here](https://github.com/GoogleCloudPlatform/microservices-demo/tree/release/v0.3.7#architecture).

#### 1. Typo
* See https://github.com/GoogleCloudPlatform/microservices-demo/issues/650.

#### 2. Public Access to Source Files
* There was a comment (from 2019) from a non-Googler (see [comment](https://github.com/GoogleCloudPlatform/microservices-demo/commit/b44d855e5c5879c80435c0891abadd3c5b246043#commitcomment-34322397)) that requested access to the source files of the diagram. Due to this interest, I've made the [Google Slides that I used to create the diagram publicly accessible](https://docs.google.com/presentation/d/12803GBApC8unpd8cmp4BOkk9dTPV4OOtihhIJHa8m28/edit?usp=sharing).

#### 3. Public Access to Source Files
* I've used a white background (for the new diagram) instead of a transparent background for developers using dark mode for GitHub. See problem with existing transparent diagram below:

Screen Shot 2022-05-13 at 9 46 21 AM

### @ahrarmonsur's Feedback
* Thank you, @ahrarmonsur, for your feedback on the diagram!
* I addressed some of it:
   * "I don't know what to look at first." - I've made the `frontend` service the focal point. I've bolded "frontend" and added a maroon/pink-ish background.
   * "The 'service' suffix might be redundant." - Suffixes removed!
   * "The edges are too attention-grabbing." - Edges lightened.
   * "Change 'Internet' to 'Client' or 'User'." - Done.
* You also suggested adding icons to each microservice. Overall, I think icons would be valuable, but I've omitted them for these reasons: 
   * The effort might not be worth it.
   * It might be hard to find well-matching icons for some of the services (e.g., `adservice`).
   * Icons might add noise to the overall diagram.
   * I don't think it's super important to communicate the purpose of each microservice.
* Let me know if I missed any other issues you raised. :)


ckim328:
this looks _clean_!! thanks for tackling the issues too!

ckim328 (approved):
This LGTM, but maybe wait for Ahrar's feedback/thoughts!

bourgeoisor (changes requested):
This looks much, much easier to read, thank you for this!
Small nit: could you make the arrow from frontend->ad start at the bottom of frontend, and not from the side?
At first glance it looks like there's an arrow from loadgen->ad which took me a while to unsee.
Similarly, could we have the arrows to checkout / payment also start from the bottom, for consistency?

NimJay:
@ckim328 Thanks for the quick review!

@bourgeoisor Good feedback! Thank you! Arrows fixed. ✅

ahrarmonsur (approved):
Wow, what a great improvement to the diagram! As the first technical artifact that informs the user of the app's architecture, it is very important to get this right. This revision is _much_ easier to consume at a glance. Great job, @NimJay!! 👏

I have some thoughts, but none of them are blocking:
- To be precise, the User/loadgenerator can communicate over both HTTP and HTTPS. In that case, the technically correct notation would be HTTP(S)
- I would suggest using the a darker shade of grey/black on the borders Service nodes (keep it consistent with the font color, too)
- Should be "Redis cache"
- The arrows from frontend to User/loadgenerator should be bi-directional (..right?)

bourgeoisor (approved):
Thank you for addressing my comment, this looks amazing!

NimJay:
Thank you, @ahrarmonsur, for the second round of feedback!

> To be precise, the User/loadgenerator can communicate over both HTTP and HTTPS. In that case, the technically correct notation would be HTTP(S)

> I would suggest using the a darker shade of grey/black on the borders Service nodes (keep it consistent with the font color, too)

> Should be "Redis cache"
> The arrows from frontend to User/loadgenerator should be bi-directional (..right?)

PR #846, opened 2022-06-09
`redis-cart` service port name

mathieu-benoit:
Following on https://github.com/GoogleCloudPlatform/microservices-demo/pull/777 to fix https://github.com/GoogleCloudPlatform/microservices-demo/issues/776

It's just a [naming convention by Istio](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/) making some features working based on this. Don't see any regression with changing this port name.

FYI, in ASM we only support `http`, `http2` and `grpc` for Service port name like explained [here](https://cloud.google.com/service-mesh/docs/naming-service-ports) to have some ASM features like the Topology graph, etc. just sharing.


mathieu-benoit:
Ready for your review, thanks.

bourgeoisor (approved):
LGTM. I'm assuming we don't call to that port by name / there's no other reference we need to change?

PR #848, opened 2022-06-09
`cartservice` - unprivileged container

mathieu-benoit:
Setup `cartservice` to run as unprivileged container

### DIFF
--- a/src/cartservice/src/Dockerfile
+++ b/src/cartservice/src/Dockerfile
@@ -16,6 +16,7 @@
 FROM mcr.microsoft.com/dotnet/sdk:6.0.300 as builder
 WORKDIR /app
 COPY cartservice.csproj .
+ENV COMPlus_EnableDiagnostics=0
 RUN dotnet restore cartservice.csproj -r linux-musl-x64
 COPY . .
 RUN dotnet publish cartservice.csproj -p:PublishSingleFile=true -r linux-musl-x64 --self-contained true -p:PublishTrimmed=True -p:TrimMode=Link -c release -o /cartservice --no-restore
@@ -28,4 +29,5 @@ RUN GRPC_HEALTH_PROBE_VERSION=v0.4.11 && \
 WORKDIR /app
 COPY --from=builder /cartservice .
 ENV ASPNETCORE_URLS http://*:7070
+ENV COMPlus_EnableDiagnostics=0
 ENTRYPOINT ["/app/cartservice"]


mathieu-benoit:
Ready for your review, thanks! Again it's just for `cartservice`, other services will come later.

The goal here is to have Online Boutique containers/apps running unprivileged (non root, etc.) in order to be easily deployed on secure environment (PSP, Gatekeeper, OpenShift, etc.). Evidence wit Bank of Anthos here: https://github.com/GoogleCloudPlatform/bank-of-anthos/issues/517. I will implement this later in there too.

src/cartservice/src/Dockerfile
```diff
@@ -28,4 +28,5 @@ RUN GRPC_HEALTH_PROBE_VERSION=v0.4.11 && \
 WORKDIR /app
 COPY --from=builder /cartservice .
 ENV ASPNETCORE_URLS http://*:7070
+ENV DOTNET_EnableDiagnostics=0
```
NimJay, line 31: I think it's fine to disable `DOTNET_EnableDiagnostics`.
I don't believe anyone working on Online Boutique uses .NET debugging and profiling (on the container itself).

kubernetes-manifests/cartservice.yaml
```diff
@@ -27,8 +27,20 @@ spec:
     spec:
       serviceAccountName: default
       terminationGracePeriodSeconds: 5
+      securityContext:
+        fsGroup: 1000
+        runAsGroup: 1000
+        runAsNonRoot: true
```
NimJay, line 33: Question: Is `runAsNonRoot` adding anything here (possibly redundant) since we already specify `runAsUser: 1000`?
I assume `runAsUser: 0` would be the `root` User (since [`runAsGroup: 0` is the root Group](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#:~:text=If%20this%20field%20is%20omitted%2C%20the%20primary%20group%20ID%20of%20the%20containers%20will%20be%20root(0).)).

NimJay (approved):
Looks good to me.
Thanks for doing this, @mathieu-benoit.
I've tested the staging URL — the cart works fine.

src/cartservice/src/Dockerfile
mathieu-benoit: Yeah, and I have read somewhere that anyway it's better for production workload to disable this anyway.

kubernetes-manifests/cartservice.yaml
mathieu-benoit: That's a good point, never thought about that. Yep agree, seems redundant, but at least more explicit and not hurting anyway? Let's keep this like this and see in the future if we want to change that. Thanks for bringing this up! Good point.

Issue #662, opened 2021-12-14
Switch to Kubernetes builtin grpc health checks


ahmetb:
I recommend doing this when (1) grpc builtin health checks are in a stable Kubernetes version (2) it's been 6 months since that version is released.

Kubernetes now supports builtin grpc health checks for probes, so we can stop using the grpc-health-probe tool I wrote for this project and further cut down external dependencies. It looks like this:

```diff
@@ -44,11 +44,15 @@ spec:
         # - name: JAEGER_SERVICE_ADDR
         #   value: "jaeger-collector:14268"
         readinessProbe:
-          exec:
-            command: ["/bin/grpc_health_probe", "-addr=:3550"]
+          grpc:
+            port: 3550
         livenessProbe:
-          exec:
-            command: ["/bin/grpc_health_probe", "-addr=:3550"]
+          grpc:
+            port: 3550
         resources:
           requests:
             cpu: 100m
```

mathieu-benoit:
This new implementation by [upstream Kubernetes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-grpc-liveness-probe) will be ideal indeed.

This feature is at the `alpha` state with Kubernetes 1.23, so it won't be yet adopted on `stable` channel yet with GKE or other managed offer ou there. [1.22 for default in `rapid` as we speak](https://cloud.google.com/kubernetes-engine/docs/release-notes).

_Note: the official doc for this feature is mentioning [`liveness probe`](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-grpc-liveness-probe) but is it working for `readiness probe` too?

_Other note: for the transition, could we have 2 `Dockerfile` files per app in order to have both container images, with or without the `grpc_health_probe` binary?_

mathieu-benoit:
Sharing the good progress of this feature in Kubernetes upstream, with the [1.24 release](https://kubernetes.io/blog/2022/05/03/kubernetes-1-24-release-announcement/):

[...]

PR #849, opened 2022-06-09
[WIP] Use upstream Kubernetes gRPC Health Check feature

mathieu-benoit:
Experimental PR to test/implement https://github.com/GoogleCloudPlatform/microservices-demo/issues/662

This won't work with the current GKE cluster used in CI because it's GKE < 1.24.

In order to test this, you could provision a GKE cluster in 1.24:
```
gcloud container clusters create test-grpc \
    --release-channel=rapid \
    --cluster-version=1.24
```
_Note: Not an option with 1.23 because it's in alpha and forces users to enable alpha features flag on their Kubernetes clusters._

- [ ] Test on GKE 1.24
- [ ] Validate approach for a potential future merge into `main`
- [ ] Run image scanning to see differences in terms of CVEs (`docker run aquasec/trivy image`)
- [ ] Update CI's + OB Live's GKE clusters to 1.24 to make sure everything works

Once tested/approved, but before thinking about merging it, what are we going to do with this PR?

If we want to support Kubernetes < 1.24 we need to:
1. either have 2 containers (one with the old bin and the other without) + some Kustomize or Helm or manifests duplication - not ideal here...
2. or wait for the end of life of 1.23 as an example which is planned for 2023-02-28 ([source](https://kubernetes.io/releases/#release-history) and [source](https://endoflife.date/kubernetes)). Apparently for GKE it's 2023-06 ([source](https://cloud.google.com/kubernetes-engine/docs/release-schedule#schedule_for_release_channels))

I think 2. is a fair option if we could say that starting 2023-02-28 (or 2023-06) Online Boutique will only work on Kubernetes 1.24+ moving forward?

Complementary information:

## Image size locally
- `emailservice`: -24%
  - `gcr.io/google-samples/microservices-demo/emailservice:v0.3.8`: 220MB
  - `gcr.io/online-boutique-ci/refs/pull/849/emailservice:849`: 190MB
- `paymentservice`: -7%
  - `gcr.io/google-samples/microservices-demo/paymentservice:v0.3.8`: 162MB
  - `gcr.io/online-boutique-ci/refs/pull/849/paymentservice:849`: 151MB
- `cartservice`: -18%
  - `gcr.io/google-samples/microservices-demo/cartservice:v0.3.8`: 62.1MB
  - `gcr.io/online-boutique-ci/refs/pull/849/cartservice:849`: 51.2MB

### DIFF
--- a/kubernetes-manifests/adservice.yaml
+++ b/kubernetes-manifests/adservice.yaml
@@ -49,15 +49,15 @@ spec:
             cpu: 300m
             memory: 300Mi
         readinessProbe:
+          grpc:
+            port: 9555
           initialDelaySeconds: 20
           periodSeconds: 15
-          exec:
-            command: ["/bin/grpc_health_probe", "-addr=:9555"]
         livenessProbe:
+          grpc:
+            port: 9555
           initialDelaySeconds: 20
           periodSeconds: 15
-          exec:
-            command: ["/bin/grpc_health_probe", "-addr=:9555"]
 ---
 apiVersion: v1
 kind: Service
--- a/kubernetes-manifests/cartservice.yaml
+++ b/kubernetes-manifests/cartservice.yaml
@@ -43,14 +43,14 @@ spec:
             cpu: 300m
             memory: 128Mi
         readinessProbe:
+          grpc:
+            port: 7070
           initialDelaySeconds: 15
-          exec:
-            command: ["/bin/grpc_health_probe", "-addr=:7070", "-rpc-timeout=5s"]
         livenessProbe:
+          grpc:
+            port: 7070
           initialDelaySeconds: 15
           periodSeconds: 10
-          exec:
-            command: ["/bin/grpc_health_probe", "-addr=:7070", "-rpc-timeout=5s"]
 ---
 apiVersion: v1
 kind: Service
--- a/kubernetes-manifests/checkoutservice.yaml
+++ b/kubernetes-manifests/checkoutservice.yaml
@@ -32,11 +32,11 @@ spec:
           ports:
           - containerPort: 5050
           readinessProbe:
-            exec:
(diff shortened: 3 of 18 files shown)


mathieu-benoit:
Yep, agreed. Option 2 seems to makes more sense.

As for the version, yep we shouldn't align this with 1.23, because the feature is in `alpha` and the Kubernetes cluster needs to enable `alpha` features, not very convenient. That's why I mentioned the alignment with 1.24.

To answer your question:
> how much value are we creating by demo-ing the feature on Online Boutique?

Top of mind, here are the reasons we should do it whenever we decide it's the right (less disruptive) time:
- Use upstream/native Kubernetes features and show best practice associated to grpc microservices
- Decrease the size and the surface of attack of the containers by integrating a bin tool in every container (for that, I will provide the before/after info about size and CVEs in the main description later)
- Avoid maintaining/updating the version of this bin embedded in every container
- Mitigate the fact that the project associated to this bin is not going to be maintained (personal assumption here) in the future (Xoogler) + this bin was doing anyway more than what we need (simple health check on the port)

NimJay:
I think just waiting for `1.24` to be default on GKE's `STABLE` channel should suffice. Like you hinted at, the 5-month footprint might be minor.
Also, thank you for listing out the value of this change. That helps!

We shall keep this PR open for now.

mathieu-benoit:
Note for myself for next rotation, let's update this PR with:
- Merge from `main` and resolve conflicts
- Update next steps based on the fact that GKE 1.25 is now in `Rapid` channel: https://cloud.google.com/kubernetes-engine/docs/release-notes#September_14_2022

mathieu-benoit:
Closing this experimental research and tests, first concrete implementation of this there now: https://github.com/GoogleCloudPlatform/microservices-demo/pull/1102

PR #906, opened 2022-06-24
Bump Python dependencies

bourgeoisor:
This PR bumps a handful of Python dependencies suggested by #728


NimJay (approved):
Thanks for doing this, @bourgeoisor!
Looks good. Approved.

PR #941, opened 2022-08-08
Remove ci check

ckim328:
### Fixes 
I added checks to the ci so that it didn't run on draft prs. Behavior was not expected, hence why we are removing it. Potentially in the future we can revisit this


bourgeoisor (approved):
LGTM

PR #959, opened 2022-08-18
Fix: Added conditional requirement for all resources in memorystore.tf

jaspermai:
### Fixes 
image

`terraform apply` was running the `kustomization-update` resource in `memorystore.tf` regardless if `memorystore = false` in `terraform.tfvars`

### Change Summary
Added a `count` attribute in the `kustomization-update` resource, that only creates an instance of the resource if `memorystore = true`.


NimJay (approved):
Thanks for addressing this quickly, Jasper!
Looks good. Approved.

ahrarmonsur (approved):
LGTM

Issue #713, opened 2022-02-08
Remove `pyyaml==6.0` as Direct Dependency


NimJay:
### Describe request or inquiry 
* This is regarding the Python services, `recommendationservice`.
* See https://github.com/GoogleCloudPlatform/microservices-demo/pull/711.
* In the above pull-request, we updated the (indirect) dependency `pyyaml` to `6.0` by adding it to the `requirements.in` file (as a "direct" dependency) and specifying the version we want (`6.0`).
* `recommendationservice` was only using `pyyaml` because of `google-python-cloud-debugger`.
* Eventually, when `google-python-cloud-debugger` upgrades its version of `pyyaml`, we will need to upgrade `google-python-cloud-debugger` and remove `pyyaml` from `requirements.in`.

### What purpose/environment will this feature serve? 
* This is a clean-up.

mathieu-benoit:
Should be fixed by https://github.com/GoogleCloudPlatform/microservices-demo/pull/985

PR #985, opened 2022-08-31
Remove `pyyaml==6.0` as Direct Dependency

mathieu-benoit:
Remove `pyyaml==6.0` as Direct Dependency, fixing https://github.com/GoogleCloudPlatform/microservices-demo/issues/713, so rollback of https://github.com/GoogleCloudPlatform/microservices-demo/pull/711 since https://github.com/GoogleCloudPlatform/microservices-demo/pull/979 has been merged into `main`.

To double check with @NimJay if it is what we need to do.


mathieu-benoit:

NimJay (approved):
All checks are passing. Looks good.
Thanks again, Mathieu!

PR #1015, opened 2022-09-01
Unpin the v4 version for `uuid`

mathieu-benoit:
In order to upgrade `uuid` to v8 with https://github.com/GoogleCloudPlatform/microservices-demo/pull/1005, we need to unpin the v4 version in the code.

See thread here talking about that: https://github.com/uuidjs/uuid/issues/594


NimJay (approved):
All checks are passing. Approved!

Issue #615, opened 2021-11-05
Missing Database in Multi-cluster with Istio


saswatmohanty01:
Original title of this issue: Cart services in multi cluster istio mode does not keep cart details while refresh browser

Cart services deployed in two Kubernetes cluster with isio has issues as there is no database , session switch between clusters and cart is empty and next refresh cart shows the correct number. Might need something with common database . Is there any fix for it.

mathieu-benoit:
What about leveraging an external service for the `redis` database? Like for example the Memorystore (redis) service? https://github.com/GoogleCloudPlatform/microservices-demo/blob/master/docs/memorystore.md

saswatmohanty01:
Thanks for your response but I am testing this on-prem which is outside GCP.

askmeegs:
Hello, if you're using Istio on-prem, could you set up multicluster service discovery and just have one instance of Redis in one cluster? And route the other cluster's cartservice to it? https://istio.io/latest/docs/setup/install/multicluster/multi-primary/#enable-endpoint-discovery

mathieu-benoit:
I just tested this scenario where the in-cluster `redis` db is in cluster 1 only and not in cluster 2. And I got the issue described here, JFYI.

mathieu-benoit:
Actually even with the cleaner and new implementation coming soon (#838) of the `cartservice` interacting with the `redis` database, the issue is still here, I just re-tested with the old and new version of `cartservice`.

And actually, that's not an issue, that's a feature :)

This behavior is due to the implementaton of the SessionId/UserId in the `frontend` app as a Cookie per instance, see here: https://github.com/GoogleCloudPlatform/microservices-demo/blob/65b552273a26714f730253a89321ac8f381882fc/src/frontend/middleware.go#L84.

Maybe a way to allow a common `redis` database and actually a common SessionId/UserId could be to have an environment variable for either the `frontend` app or the `cartservie` app. This environment variable could be for example named `UNIQUE_USER_ID_ACCROSS_SESSIONS` or something like this, and then either the `frontend` or the `cartservice` could use a unique and comment value (for example `UUI.Empty` or equivalent). Sharing my thoughts and findings for potential future implementation around this.

mathieu-benoit:
These 2 PRS https://github.com/GoogleCloudPlatform/microservices-demo/pull/838 and https://github.com/GoogleCloudPlatform/microservices-demo/pull/1023 allow to support this scenario now.

PR #1023, opened 2022-09-02
Allow to have a unique `user-id` across sessions

mathieu-benoit:
This introduces a new environment variable for the `frontend` app: `ENABLE_SINGLE_SHARED_SESSION`. This allows to have a unique User Id across the sessions. The default behavior is one User Id per session.

Fixing https://github.com/GoogleCloudPlatform/microservices-demo/issues/615 + allows more scenarios to showcase:
- Multi-clusters Mesh setup ([like here with BoA](https://github.com/GoogleCloudPlatform/anthos-service-mesh-samples/tree/main/demos/asm-mcp-multi-clusters)) where the in-cluster `redis` database is just in 1 cluster, see that the data are accessible from `cartservice` and `frontend` from different clusters.
- Multi-clusters (HA on different regions for example) talking to one Memorystore (redis) instance, see that the data are accessible from `cartservice` and `frontend` from different clusters.
- Same setup than previously but soon with Spanner ;)

If you set this environment variable to `"true"` on the `frontend` app, here is how to test this new behavior:
- if you open your internet browser you will see the `user-id` at the bottom of the website (associated cookie) with the value: ``. If you open your internet browser now in InPrivate/Incognito mode, you will see the same value.
- if you try to add items in your shopping cart, you will see the same items in your shopping cart accross your InPrivate/Incognito mode or the other.

_Notes:_
- _This implementation with this PR is not updating the Kubernetes manifest, if someone wants to leverage this feature they could edit the `frontend`'s `Deployment` manifest or for example run this command: `kubectl set env deployment/frontend ENABLE_SINGLE_SHARED_SESSION="true"`._
- _No doc update either yet._
- _Later, we may want to have a Kustomize flavor/overlay with that._


src/frontend/middleware.go
```diff
@@ -86,8 +87,13 @@ func ensureSessionID(next http.Handler) http.HandlerFunc {
...
 		if err == http.ErrNoCookie {
-			u, _ := uuid.NewRandom()
-			sessionID = u.String()
+			if os.Getenv("ENABLE_UNIQUE_USER_ID_ACROSS_SESSIONS") == "" {
+				// Hard coded user id, shared across sessions
+				sessionID = "12345678-1234-1234-1234-123456789123"
+			} else {
+				u, _ := uuid.NewRandom()
+				sessionID = u.String()
+			}
```
NimJay, line 96: Nitpick (feel free to ignore):
We might be able to shorten this chunk of code.
```
sessionID = "12345678-1234-1234-1234-123456789123"
if os.Getenv("ENABLE_UNIQUE_USER_ID_ACROSS_SESSIONS") != "" {
	u, _ := uuid.NewRandom()
	sessionID = u.String()
}
```

mathieu-benoit: Yeah, I thought about something like this, but with this proposition we read that the default behavior/value is "a unique id across sessions" which is not the case. I think I would prefer the `if... else...` to clarify that point.

NimJay: **Question:** 
Is there a reason the default behaviour is having a single, shared session? Do you think it's more useful?
I think `ENABLE_UNIQUE_USER_ID_ACROSS_SESSIONS = true` should be the default behaviour because:
1. during live demos (or even on [onlineboutique.dev](https://onlineboutique.dev/)), each visitor should have their own cart (to avoid confusion)
2. it's what uses are already familiar with (since it's the _current_ behaviour)

**Question 2:**
In which, case should we use a different environment variable name such as `ENABLE_SINGLE_SHARED_SESSION` (where `ENABLE_SINGLE_SHARED_SESSION` means the opposite of `ENABLE_UNIQUE_USER_ID_ACROSS_SESSIONS`)?

mathieu-benoit: Yep, like I mentioned in the description:

> FIXME: before reviewing/merging, the test is with the value "" for a testing purpose, but it should be "true".

Sorry if it was not clear, for now with `""` it's just to test quickly, then I will set the proper value `"true"` in there ;)

Sorry for the confusion at this stage.

mathieu-benoit: RE: `ENABLE_SINGLE_SHARED_SESSION`, I'm not sure I get what you mean, could you elaborate?

NimJay: Ohhh, got it. What I understand: you'll be adding `ENABLE_UNIQUE_USER_ID_ACROSS_SESSIONS: true` to the `/kubernetes-manifests` (before merging this PR).
**Suggestion:** If that's the case, I suggest using an environment variable that means the opposite such as `ENABLE_SINGLE_SHARED_SESSION` (so that by default, we don't have to think about this variable at all).

mathieu-benoit: Sorry, nop, I'm having the test with `true` in the code directly, actually I just did that commit https://github.com/GoogleCloudPlatform/microservices-demo/pull/1023/commits/d236d8d736e2e27e31a4b9747cb2fcb1cfa8acde.

Does that make more sense now with this commit?

mathieu-benoit:
Just did my tests, ready for your review, thanks!

Sorry for the confusion while I was doing my test in draft mode, if there is still confusion or your need clarification, feel free to ping me.

The default behavior is still the same, now if someone wants a unique user id across sessions, they could use this env var `ENABLE_UNIQUE_USER_ID_ACROSS_SESSIONS = true` on the `frontend`'s `Deployment`.

src/frontend/middleware.go
NimJay: Yup, just saw you commit. Thanks for that quick clarification!

I'll clarify what I mean by `ENABLE_SINGLE_SHARED_SESSION`:
* `ENABLE_SINGLE_SHARED_SESSION` means the opposite of `ENABLE_UNIQUE_USER_ID_ACROSS_SESSIONS`
* Right now, the default deployment of Online Boutique requires `ENABLE_UNIQUE_USER_ID_ACROSS_SESSIONS` to be set to `true`
* If, instead of `ENABLE_UNIQUE_USER_ID_ACROSS_SESSIONS`, we introduce `ENABLE_SINGLE_SHARED_SESSION` in this PR, we won't have to touch the `kubernetes-manifests/` and we could just introduce a new Kustomize Component similar [/kustomize/components/cymbal-branding](https://github.com/GoogleCloudPlatform/microservices-demo/tree/main/kustomize/components/cymbal-branding).
* Also, if we use `ENABLE_SINGLE_SHARED_SESSION`, the code could be:
```
u, _ := uuid.NewRandom()
sessionID = u.String()
if os.Getenv("ENABLE_SINGLE_SHARED_SESSION") != "" {
	sessionID = "12345678-1234-1234-1234-123456789123" // Hard code the single session shared by all visitors
}
```
or something similar.

But this is a minor detail, so I don't feel strongly.

mathieu-benoit: > Right now, the default deployment of Online Boutique requires ENABLE_UNIQUE_USER_ID_ACROSS_SESSIONS to be set to true

True, if you want the new behavior ("a unique user id accross sessions").

But not true, if you want to keep the default behavior ("new user id per session").

mathieu-benoit: Like discussed in 1:1, let's do `ENABLE_SINGLE_SHARED_SESSION`.

NimJay (approved):

PR #887, opened 2022-06-20
Unprivileged containers

mathieu-benoit:

PR #1046, opened 2022-09-14
`loadgenerator` - `securityContext` for `initContainer`

mathieu-benoit:
`loadgenerator` - `securityContext` for `initContainer`

Was missing when did this https://github.com/GoogleCloudPlatform/microservices-demo/pull/887

Got raised when applying PoCo's `Constraints`:
```
psp-capabilities | init container <frontend-check> is not dropping all required capabilities. Container must drop all of ["NET_RAW"] or "ALL"
psp-allow-privilege-escalation-container | Privilege escalation container is not allowed: frontend-check
```

### DIFF
--- a/kubernetes-manifests/loadgenerator.yaml
+++ b/kubernetes-manifests/loadgenerator.yaml
@@ -47,6 +47,13 @@ spec:
               exit 1
           fi
         name: frontend-check
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+              - all
+          privileged: false
+          readOnlyRootFilesystem: true
         image: busybox:latest
         env:
         - name: FRONTEND_ADDR


mathieu-benoit:

NimJay (approved):
Good catch! I totally forgot about the `loadgenerator`'s `initContainer`.
Approved!

**Just a quick question:** what is `PoCo` ~and `psp-capabilities`~?
**Update:** Ah, "psp" stands for "Pod Security Policy". Got it. But what is `PoCo`?

mathieu-benoit:
> Just a quick question: what is `PoCo` and `psp-capabilities`?

Policy Controller's `Constraint`: https://cloud.google.com/anthos-config-management/docs/latest/reference/constraint-template-library#k8spspcapabilities ;)

Issue #712, opened 2022-02-08
Remove `semistandard`


NimJay:
### Describe request or inquiry 
* Both Node.js services import [semistandard](https://github.com/standard/semistandard) as an npm `devDependecy`.
* But I don't think any developers of this repo actually uses it.
* Services affected:
  * currencyservice
https://github.com/GoogleCloudPlatform/microservices-demo/blob/d8936827909bcf18d554dce5064671f9465e115e/src/currencyservice/package.json#L24
  * paymentservice
https://github.com/GoogleCloudPlatform/microservices-demo/blob/d8936827909bcf18d554dce5064671f9465e115e/src/paymentservice/package.json#L24
* We need to:
* [ ] Double check that [semistandard](https://github.com/standard/semistandard) isn't being used in a meaningful way.
* [ ] Delete [semistandard](https://github.com/standard/semistandard) usage.

### What purpose/environment will this feature serve? 
* This is a clean-up task.
* This will help reduce the bloat of this app.

### Additional Info
* This is related to https://github.com/GoogleCloudPlatform/microservices-demo/issues/675.

bourgeoisor:
I think we should still do this. Commenting to reset SLO.

PR #1051, opened 2022-09-14
adressing Remove semistandard #712

smeet07:
### Background 
I don't think semistandard is being used in a meaningful way therefore removed it as a devdependency

### Fixes 
Remove semistandard #712

### DIFF
--- a/src/currencyservice/package.json
+++ b/src/currencyservice/package.json
@@ -5,7 +5,6 @@
   "repository": "https://github.com/GoogleCloudPlatform/microservices-demo",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "lint": "semistandard *.js"
   },
   "license": "Apache-2.0",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


src/currencyservice/package.json
```diff
@@ -20,7 +20,4 @@
     "request": "2.88.2",
     "xml2js": "0.4.23"
   },
-  "devDependencies": {
```
mathieu-benoit, line 23: I also see `"lint": "semistandard *.js"` line 8, does this need to be removed too?

bourgeoisor: I'd assume so if we're removing the module entirely

smeet07:
yes I think we could remove that as well

smeet07:
since we are removing semistandard module, the script wont work anyways

bourgeoisor (approved):
LGTM

mathieu-benoit:
Thanks for your contribution @smeet07, much appreciated!

smeet07:
No problem, Always happy to help

PR #1096, opened 2022-09-26
Add a small Terraform CI

mathieu-benoit:
Add a small Terraform CI to at least run `terraform validate` as soon as a change is made in the `terraform/` folder.

Also, added an `ignore-paths` filtering for the current CI definition triggered at any PR or merge into `main`, because there is no relationship at all (as for now at least) with the `terraform/` folder. So to avoid any `self-hosted` runner to be used for this and also to avoid any container images to be built for that. So if only Terraform updates are made (Terraform developer, Renovate, etc.) no build nor "PR staged" anymore.

Also, not using the `self-hosted` runner, just use the prebuilt GitHub runner for this, smaller footprint for what we want to achieve with this `tf-validate-ci` workflow (actually not deploying anything in GCP).

Finally, I renamed `ci-master.yaml` to `ci-main.yaml` for naming convention.


ckim328 (approved):

bourgeoisor (approved):
I like this idea -- this looks great!

.github/workflows/terraform-validate-ci.yaml
```diff
@@ -0,0 +1,36 @@
...
+    paths:
+      - 'terraform/**'
+      - '.github/workflows/terraform-validate-ci.yaml'
+  pull_request:
+    paths:
+      - 'terraform/**'
+      - '.github/workflows/terraform-validate-ci.yaml'
+jobs:
+  job:
+    runs-on: ubuntu-22.04
```
NimJay, line 29: Praise: I like that we're using GitHub-hosted VMs! Maintenance of `self-hosted` virtual machines is not great.

.github/workflows/terraform-validate-ci.yaml
```diff
@@ -0,0 +1,36 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: terraform init & validate
+        run: |
+          cd terraform/
+          terraform init -backend=false
```
NimJay, line 35: Praise: Thanks for adding `-backend=false`.
Just for my understanding, will this be useful if we ever choose to use something like Google Cloud Storage to maintain Terraform's state, etc?

.github/workflows/ci-pr.yaml
```diff
@@ -17,6 +17,10 @@ on:
   pull_request:
     branches:
       - main
+    paths-ignore:
+      - '**/README.md'
+      - 'terraform/**'
+      - '.github/workflows/terraform-validate-ci.yaml'
```
NimJay, line 23: Question:
Is there a reason we don't add this to [/.github/workflows/ci-master.yaml](https://github.com/GoogleCloudPlatform/microservices-demo/blob/43b7ef46d90ff8bba7d0f2681792db4447232dd0/.github/workflows/ci-master.yaml)?

.github/workflows/terraform-validate-ci.yaml
mathieu-benoit: For this `terraform init` step, not that much, in case we want to make sure that this step is working with GCS. Other than that, if we start having an interaction with GCP, it could allow to go step further in our tests with `terraform plan` which requires to be authenticated to GCP. Does that answer your question?

NimJay: Ahh, that makes sense. Thanks, Mathieu!

NimJay (approved):
I left one question.
I also made a minor change to a file name: `tf-validate-ci.yml` is now `terraform-validate-ci.yaml`.
* `.yaml` for consistency.
* `terraform-` for clarity.

.github/workflows/ci-pr.yaml
mathieu-benoit: Good point/catch! I thought I had a reason, but actually that totally makes sense to have this in there too! Working on that.

mathieu-benoit: Just took into account this. Note that I also renamed `ci-master.yaml` to` by `ci-main.yaml`, just heads up.

PR #1102, opened 2022-09-27
Kustomize variation to deploy Online Boutique apps without the `grpc-health-probe` bin

mathieu-benoit:
- Support [native gRPC Healthcheck for `livenessProbe` and `readinessProbe` since GKE 1.24](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-grpc-liveness-probe).
- Partially implementing #662 (following up on first experimental tests documented here #849)
  - "Partially" because it won't be yet by default, not all GKE versions support that, just 1.24+. So it will be an optional Kustomize variation that users would be able to leverage if they want.
- Size of the images got -3.9MB (virtual) / -10.9MB (on disk)
- `loadgenerator` and `frontend` are not impacted by that (not gRPC apis/apps)
- Adding 3 more Kustomize variations related to this one:
  - `container-images-registry` (not needed, but similar to the one below)
  - `container-images-tag` (not needed, but similar to the one below)
  - `container-images-tag-suffix` (needed for this new Kustomize variation)
- Review the setup of tokens to avoid issues with Kustomize and bash session
- Update the release process (more precisely `make-docker-images.sh`) in order to build a second image next to the default/official one by adding a prefix `-native-grpc-probes`.
- These `-native-grpc-probes` images are not built at every PR nor merge into `main`, they are only built during the release process.
- Added those new Kustomize variations in the Smoke test in CI, with some `sed` examples too

Tests realized on a GKE 1.24 now available on `rapid` channel:
```
gcloud container clusters create tests \
    --zone=us-central1-b \
    --machine-type=e2-standard-2 \
    --num-nodes=4 \
    --release-channel=rapid
```
Build the containers for all the Online Boutique apps with the `-native-grpc-probes` image tag, you can leverage the script [here](https://github.com/GoogleCloudPlatform/microservices-demo/blob/mathieu-benoit/without-grpc-health-probe-bin/hack/make-docker-images.sh).

Then, deploy this like:
```
cd kustomize/
SUFFIX=-native-grpc-probes
sed -i "s/CONTAINER_IMAGES_TAG_SUFFIX/$SUFFIX/g" components/container-images-tag-suffix/kustomization.yaml
kustomize edit add component components/container-images-tag-suffix
kustomize edit add component components/native-grpc-health-check
kubectl kustomize . | sed "s/$SUFFIX$SUFFIX/$SUFFIX/g" | kubectl apply -f
```
_Note: for this variation, `kubectl apply -k .` won't work because there is a [known issue currently in Kustomize](https://github.com/kubernetes-sigs/kustomize/issues/4814) where the `tagSuffix` is duplicated. The command lines above are a temporary workaround._

We won't make this the default configuration yet, GKE 1.24 is the default version for the `rapid` channel. We need to make the choice in the future when this configuration will be the default ones. [GKE just got 1.24 in rapid channel in 2022-08-12](https://cloud.google.com/kubernetes-engine/docs/release-schedule) and is planned to have its end of life on 2023-09-30. So it's 1 year from now. Will it be a good time then, after 2023-09-30?


mathieu-benoit:

hack/make-docker-images.sh
```diff
@@ -38,9 +38,11 @@ while IFS= read -d $'\0' -r dir; do
         cd "${builddir}"
         log "Building: ${image}"
         docker build -t "${image}" .
+        docker build -t "${image}-native-grpc-probes" . --target without-grpc-health-probe-bin
 
         log "Pushing: ${image}"
         docker push "${image}"
+        docker push "${image}-native-grpc-probes"
```
NimJay, line 45: Praise: Thanks for building and pushing the new images in the release scripts (`/hack/make-release.sh`).

kustomize/components/memorystore/kustomization.yaml
```diff
@@ -28,7 +28,7 @@ patchesStrategicMerge:
           - name: server
             env:
             - name: REDIS_ADDR
-              value: "{{REDIS_ADDR}}"
+              value: "REDIS_ADDR"
```
NimJay, line 31: Question: Is there a reason we remove the curly braces? 
Should we also fix:
```
sed -i "s/{{REDIS_ADDR}}/${REDIS_IP}:${REDIS_PORT}/g" components/memorystore/kustomization.yaml
```
inside `kustomize/components/memorystore/README.md`?

NimJay: Maybe it was an accident?
```suggestion
              value: "{{REDIS_ADDR}}"
```

kustomize/components/container-images-registry/README.md
```diff
@@ -0,0 +1,28 @@
...
+By default, Online Boutique's services' container images are pulled from a public container registry (`gcr.io/google-samples/microservices-demo`). One best practice is to have these container images in your own private container registry. The Kustomize variation in this folder can help with using your own private container registry.
+
+## Change the default container registry via Kustomize
+
+To automate the deployment of Online Boutique integrated with your own container registry, you can leverage the following variation with [Kustomize](../..).
+
+From the `kustomize/` folder at the root level of this repository, execute this command:
+```
+REGISTRY=my-registry # Example: gcr.io/my-project/my-directory
+sed -i "s|CONTAINER_IMAGES_REGISTRY|${REGISTRY}|g" components/container-images-registry/kustomization.yaml
```
NimJay, line 12: **Rationale for my change:**
I've changed `"s/CONTAINER_IMAGES_REGISTRY/$REGISTRY/g"` to `"s|CONTAINER_IMAGES_REGISTRY|$REGISTRY|g"` because the `/`s inside `gcr.io/my-project/my-directory` was breaking the `sed` command. So we're no longer using regexes (using `|`s) to avoid `/`s being interpreted as special characters.

kustomize/components/memorystore/kustomization.yaml
mathieu-benoit: Hey Nim, good question, actually I put that in the description: `Review the setup of tokens to avoid issues with Kustomize and bash session`. So that's on purpose otherwise we have issue in bash session when doing the `sed` part and/or when doing `kustomize build .` without these tokens replaced. Let me know if you need more info. But the tl,dr is we should have token just like this `REDIS_ADDR` not with any `{}` nor `{{}}`, etc.

kustomize/components/container-images-tag-suffix/README.md
```diff
@@ -0,0 +1,46 @@
...
+kind: Kustomization
+resources:
+- base
+components:
+- components/container-images-tag-suffix
+```
+
+You can locally render these manifests by running `kubectl kustomize . | sed "s/$SUFFIX$SUFFIX/$SUFFIX/g"` as well as deploying them by running `kubectl kustomize . | sed "s/$SUFFIX$SUFFIX/$SUFFIX/g" | kubectl apply -f`.
+
+_Note: for this variation, `kubectl apply -k .` alone won't work because there is a [known issue currently in Kustomize](https://github.com/kubernetes-sigs/kustomize/issues/4814) where the `tagSuffix` is duplicated. The `sed "s/$SUFFIX$SUFFIX/$SUFFIX/g"` commands above are a temporary workaround._
```
NimJay, line 30: Praise: Thank you for
1. catching this issue with the duplicated `tagSuffix`
2. adding a workaround
3. adding a note about the workaround
This is great!

kustomize/components/memorystore/README.md
```diff
@@ -47,7 +47,7 @@ components:
 ```
 
 Update current Kustomize manifest to target this Memorystore (Redis) instance.
-```bash
+```sh
 REDIS_IP=$(gcloud redis instances describe redis-cart --region=${REGION} --format='get(host)')
 REDIS_PORT=$(gcloud redis instances describe redis-cart --region=${REGION} --format='get(port)')
 sed -i "s/{{REDIS_ADDR}}/${REDIS_IP}:${REDIS_PORT}/g" components/memorystore/kustomization.yaml
```
NimJay, line 53: **Suggestion:** Should we update this `sed` command?
```suggestion
sed -i "s/REDIS_ADDR/${REDIS_IP}:${REDIS_PORT}/g" components/memorystore/kustomization.yaml
```
Regarding: https://github.com/GoogleCloudPlatform/microservices-demo/pull/1102#discussion_r1001900094

mathieu-benoit: Oh yeah, maybe a merge issue, let me check that and update this.

kustomize/kustomization.yaml
```diff
@@ -23,3 +23,7 @@ components:
 # - components/network-policies
 # - components/service-accounts
 # - components/spanner
+# - components/container-images-tag
+# - components/container-images-tag-suffix
+# - components/container-images-registry
```
NimJay, line 28: **Praise:** 
Thanks for putting these in the working right order (instead of in alphabetical order)!
```
- components/container-images-tag
- components/container-images-tag-suffix
- components/container-images-registry
```

NimJay (approved):
Thanks again, Mathieu, for proactively figuring (and implementing) this out! 👏 
Everything looks good.

I have tested all the Kustomize building.
I've committed some minor fixes/edits to avoid back-and-forth.

But I have not deployed the new `-native-grpc-probes` to a GKE cluster, but I trust that they work.

Feel free to merge whenever.

kustomize/components/memorystore/kustomization.yaml
LukeSchlangen: Following this today, one possible unintended side effect is that the `sed` now matches the name and value. So I end up with:

```
            - name: 10.34.24.75:6379
              value: "10.34.24.75:6379"
```

mathieu-benoit: Yep, that's why we document this like this: https://github.com/GoogleCloudPlatform/microservices-demo/pull/1102/commits/f60a67fabbd855df4c0f703ea6bb3918e4b214f1. But yeah if someone is not doing the `\"` it could land to error. Open for suggestions to improve this for the different files having tokens like this. PR about that are more than welcome ;)

mathieu-benoit: Note: utlimately, I'm wondering if we have something to do with Kustomize's [`replacements`](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/replacements/) (knowing that [`vars`](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/vars/) are planned to be deprecated).

LukeSchlangen: Interesting. I didn't run the command manually, I was following the [terraform](https://github.com/GoogleCloudPlatform/microservices-demo/tree/main/terraform) example when that happened. I'll try running it again to see if I get the same result?

mathieu-benoit: Yes please, if you run into any issue, feel free to file an Issue/PR for that. For Terraform, there is this https://github.com/GoogleCloudPlatform/microservices-demo/blob/main/terraform/memorystore.tf#L37, JFYI. Which is not working with what you just said?

LukeSchlangen: Correct. I checked that I pulled latest and just reran the terraform deploy. I see the escape characters and the quote `\"` and it still seems to be replacing all of them.

To clarify, it worked in the sense that the application deployed beautifully, it just replaced all instances of `REDIS_ADDR` with `10.34.24.75:6379`. (Including the instance at the top of the `/microservices-demo/kustomize/components/memorystore/kustomization.yaml` file.)

```diff
-# cartservice - replace REDIS_ADDR to target new Memorystore (redis) instance
+# cartservice - replace 10.34.24.75:6379 to target new Memorystore (redis) instance
 - |-
   apiVersion: apps/v1
   kind: Deployment
@@ -27,8 +27,8 @@ patchesStrategicMerge:
         containers:
           - name: server
             env:
-            - name: REDIS_ADDR
-              value: "REDIS_ADDR"
+            - name: 10.34.24.75:6379
+              value: "10.34.24.75:6379"
 # redis - remove the redis-cart Deployment
 - |-
   apiVersion: apps/v1
```

mathieu-benoit: Gotcha, thanks. Do you mind reporting this as a dedicated issue please? We will need to tackle this to fix this broken Terraform deployment with Memorystore. Thanks!

LukeSchlangen: I'll do that!

mathieu-benoit: FYI: this will be fixed via https://github.com/GoogleCloudPlatform/microservices-demo/pull/1202. Thanks!

LukeSchlangen: Nice!

I also had a solution (all the backslashes!!) that I was going to PR today, but that is much cleaner: https://github.com/LukeSchlangen/microservices-demo/commit/4f1b278111782cfcdda4224d77fc501d10ada8ed

PR #1111, opened 2022-09-30
paths-ignore for cleanup ns

mathieu-benoit:
`paths-ignore` for cleanup `Namespaces`


NimJay (approved):
Good catch! Approved.

PR #1143, opened 2022-10-06
`redis-cart`'s port name update from `tls-redis` --> `tcp-redis`

mathieu-benoit:
When I did https://github.com/GoogleCloudPlatform/microservices-demo/pull/846, I actually did it wrong with a typo `tls-` instead of `tcp-`, fixing this now :)


NimJay (approved):
Thanks for catching and fixing this, Mathieu!

PR #1167, opened 2022-10-14
feat!: Add OpenTelemetry Collector

arbrown:
* Add OpenTelemetry Collector and expose it as a service
* Instrument frontend to send traces for http requests to the collector
* Change behavior of environment variables configuring tracing and stats for frontend


kubernetes-manifests/collector.yaml
```diff
@@ -0,0 +1,107 @@
...
+      # and inserts it into the collector config template
+      # https://cloud.google.com/compute/docs/storing-retrieving-metadata
+      initContainers:
+      - name: otel-gateway-init
+        image: gcr.io/google.com/cloudsdktool/cloud-sdk:405.0.0-alpine
+        command:
+        - '/bin/bash'
+        - '-c'
+        - |
+           sed "s/{{PROJECT_ID}}/$(curl -H 'Metadata-Flavor: Google' http://metadata.google.internal/computeMetadata/v1/project/project-id)/" /template/collector-gateway-config-template.yaml >> /conf/collector-gateway-config.yaml
```
NimJay, line 40: **Question for our 1-on-1:**
What is the purpose of this init container?

kubernetes-manifests/collector.yaml
```diff
@@ -0,0 +1,107 @@
...
+    service:
+      pipelines:
+        traces:
+          receivers: [otlp] # Receive otlp-formatted data from other collector instances
+          processors: []
+          exporters: [googlecloud] # Export traces directly to Google Cloud
+        metrics:
+          receivers: [otlp]
+          processors: []
+          exporters: [googlecloud] # Export metrics to Google Cloud
```
NimJay, line 107: **Question for our 1-on-1:**
Thoughts on turning this into a Kustomize component (see [relevant Google Doc comment](https://docs.google.com/document/d/1Px8ZIDaLc8Z4YE4w8iGujNWPmrPxj42NioEPTC0CS5E/edit?resourcekey=0-zYlIITSyKkh7PVUypWleVw&disco=AAAAg04Qsp4))?

arbrown: Done. PTAL. I also added a local-images components to require using a local (to a project) image of each service for development purposes.  This addresses the issue we discussed earlier about testing kustomize-specific deployments like the collector using skaffold.

skaffold.yaml
```diff
@@ -49,19 +49,8 @@ build:
...
-    - ./kubernetes-manifests/currencyservice.yaml
-    - ./kubernetes-manifests/emailservice.yaml
-    - ./kubernetes-manifests/frontend.yaml
-    - ./kubernetes-manifests/paymentservice.yaml
-    - ./kubernetes-manifests/productcatalogservice.yaml
-    - ./kubernetes-manifests/recommendationservice.yaml
-    - ./kubernetes-manifests/redis.yaml
-    - ./kubernetes-manifests/shippingservice.yaml
+  kustomize:
+    paths: [kustomize]
```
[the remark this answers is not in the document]
arbrown: Nim and I discussed this and I said I'd look into the possibility, but I haven't had a chance to go over the implementation with him.  I think this is a positive change as it keeps the skaffold deployment in line with the one we use with kustomize.  This is important now that we are using kustomize modules to do more than just change some configs.  This seems like the best (only?) way to use skaffold for testing kustomize modules that add new deployments like this collector.  Thoughts?

kustomize/components/google-cloud-operations/otel-collector.yaml
```diff
@@ -0,0 +1,107 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+apiVersion: apps/v1
+kind: Deployment
```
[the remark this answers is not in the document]
arbrown: Done.  It seems to work correctly, but PTAL.

skaffold.yaml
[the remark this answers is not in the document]
NimJay: **Suggestion**:
Could we undo this `skaffold.yaml` change in this pull-request?
I think there's a larger discussion to be had here and should be done in a separate pull-request. I would also like to test the development workflow because merging this `skaffold` change.
Reason we to discuss further: We don't want `kustomize/` to contain development manifests since a lot of demos and tutorials use the `main` branch (e.g., [this](https://cloud.google.com/service-mesh/docs/by-example/canary-deployment)). Reminder: The `kustomize/base/` samples are updated every time a new release is created.

NimJay (approved):
I have not tested thoroughly (e.g., in my own cluster). But I trust you've done the appropriate testing. Plus, all the checks are passing so we know that the images build. Approved.

kustomize/components/google-cloud-operations/otel-collector.yaml
```diff
@@ -0,0 +1,126 @@
...
+      initContainers:
+      - name: otel-gateway-init
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+              - all
+          privileged: false
+          readOnlyRootFilesystem: true
+        image: gcr.io/google.com/cloudsdktool/cloud-sdk:405.0.0-alpine
```
mathieu-benoit, line 47: Is the `gcr.io/google.com/cloudsdktool/cloud-sdk:405.0.0-alpine` image mandatory here?

In `loadgenerator` we use `busybox` (https://github.com/GoogleCloudPlatform/microservices-demo/blob/main/kubernetes-manifests/loadgenerator.yaml#L57), could use the same one here too?

mathieu-benoit (dismissed):

skaffold.yaml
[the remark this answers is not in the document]
arbrown: I'm not sure I understand your comments here and I think this warrants further discussion.  I'm happy to pull this skaffold change out but only _because_ I've already used it to test the changes and I'm confident in them.

The core of the issue (as I see it) is that we need some way to `skaffold run` the same configurations we're putting in `kustomize/`.  This is especially true for modules like this one that include new deployments and not just config changes.  There might be a better solution than the one I have been using, but it is aligned with our current recommendation of using kustomize by commenting/uncommenting modules in `kustomize/kustomization.yaml`.  This should not interfere with any demos and tutorials since it does not change any of the manifest files and the default behavior will be to follow the default kustomize deployment (without any changes.)

We can remove these changes, but we'll need _some_ way of testing kustomize changes, and I think that maintaining parallel configs (one in kustomize, one in skaffold) for every combination of modules is not feasible.

kustomize/components/google-cloud-operations/otel-collector.yaml
arbrown: Yes.  Updated.

kustomize/components/google-cloud-operations/otel-collector.yaml
```diff
@@ -37,16 +37,9 @@ spec:
       # https://cloud.google.com/compute/docs/storing-retrieving-metadata
       initContainers:
       - name: otel-gateway-init
-        securityContext:
-          allowPrivilegeEscalation: false
-          capabilities:
-            drop:
-              - all
-          privileged: false
-          readOnlyRootFilesystem: true
```
NimJay: **Question:**
Is there a reason we removed this:
```
        securityContext:
          allowPrivilegeEscalation: false
          capabilities:
            drop:
              - all
          privileged: false
          readOnlyRootFilesystem: true
```

skaffold.yaml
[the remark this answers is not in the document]
NimJay: I 100% agree that we should `skaffold`-ify our development manifests.
And thank you so much for looking into this. I think the solution that you started is clever (e.g., having a new Kustomize Component for development).

But let's take this to another pull-request since it's somewhat unrelated to OpenTel, and because there are some details we need to flesh out such as:
1. could we use Kustomize's [`images`](https://github.com/NimJay/intro-to-kustomize/blob/main/images/kustomization.yaml) transformer or something similar?
1. how should we update the doc inside [`/docs/development-guide.md`](https://github.com/GoogleCloudPlatform/microservices-demo/blob/main/docs/development-guide.md)?
1. should we delete `/kubernetes-manifests/*`?
1. how should the release automation inside `hack/` be updated?
1. how will this effect people using the `main` branch in their demos/tutorials? Should we introduce a new `git` tag or branch `latest` that points to the latest release of this repository?

I would also love to flesh out some of these details via a Google Meet. 
What you think, Drew?

NimJay:
@arbrown, apologies for the delayed review, and thank you for addressing all my concerns!
⚠️: Let's pull away the changes related to `skaffold` for now and drop that into another pull-request. See [my other comment](https://github.com/GoogleCloudPlatform/microservices-demo/pull/1167#discussion_r1010486249) for more info.

Everything else looks good!
Feel free to merge once ⚠️ has been done.

kustomize/components/google-cloud-operations/otel-collector.yaml
arbrown: I think that's a victim of a merge gone wrong.  I've fixed it.

PR #1204, opened 2022-10-23
`adservice` - `eclipse-temurin:18.0.2.1_1-jre-alpine`

mathieu-benoit:
Doesn't change anything yet with the current version because `eclipse-temurin:18-jre-alpine` and `eclipse-temurin:18.0.2.1_1-jre-alpine` are similar since 2 months. But it's following up on https://github.com/GoogleCloudPlatform/microservices-demo/pull/1168 in order to better track/catch CVEs vulns fix in the future.


bourgeoisor (approved):
LGTM

PR #1205, opened 2022-10-23
Force `git pull` in the release process

mathieu-benoit:
Proposing to force `git pull` in the release process.

It looks like `v0.4.1` was not generated from the latest and greatest from the `main` branch.

At least for the `frontend` app. I found out that `v0.4.1` for `frontend` still has 2 vulnerabilities with a known fix:
```
CVE-2022-40304 | Unspecified | 0 | Yes | libxml2 | OS
CVE-2022-40303 | Unspecified | 0 | Yes | libxml2 | OS
```
But if I do:
```
git checkout main
git pull
cd src/frontend
docker build...
docker push...
```
I don't see these 2 CVE vulnerabilities anymore. So guessing the issue came from this.

Once this PR is merged in the `main` branch, please let's generate `v0.4.2` release of Online Boutique with this (the GKE Security Posture feature will be happy with that :)). Thanks!

### DIFF
--- a/hack/make-release.sh
+++ b/hack/make-release.sh
@@ -34,6 +34,10 @@ if [[ "$TAG" != v* ]]; then
     fail "\$TAG must start with 'v', e.g. v0.1.0 (got: $TAG)"
 fi
 
+# make sure local source is up to date
+git checkout main
+git pull
+
 # build and push images
 "${SCRIPTDIR}"/make-docker-images.sh
 
--- a/hack/make-release.sh
+++ b/hack/make-release.sh
@@ -34,6 +34,12 @@ if [[ "$TAG" != v* ]]; then
     fail "\$TAG must start with 'v', e.g. v0.1.0 (got: $TAG)"
 fi
 
+# ensure there are no uncommitted changes
+if [[ $(git status -s | wc -l) -gt 0 ]]; then
+    echo "error: can't have uncommitted changes"
+    exit 1
+fi
+
 # make sure local source is up to date
 git checkout main
 git pull


hack/make-release.sh
```diff
@@ -34,6 +34,10 @@ if [[ "$TAG" != v* ]]; then
     fail "\$TAG must start with 'v', e.g. v0.1.0 (got: $TAG)"
 fi
 
+# make sure local source is up to date
+git checkout main
+git pull
```
bourgeoisor, line 39: Instead of silently changing branch (if applicable) and pulling (if applicable), may I suggest we loudly fail instead?

e.g. for the branch
```
if [[ <branch is not main> ]]; then
    echo "error: need to be checked in on the main branch"
    exit 1
fi
```

Could also have `if [[ <branch not in sync with upstream> ]]; then` etc.

That way the "user" of the release script realize before continuing with the script that "ah, hold on, what is happening here, am I in the wrong place?" and have the time to double check some more stuff before proceeding with the rest of the releasing script.

mathieu-benoit: I see what you mean, but here the goal is to make sure it could be automated, without any user interaction. Ultimately, this script needs to be executed by a fully automated process, no local script execution. So this `git checkout` + `git pull` should mimic this part, don't you think?

bourgeoisor: I see what you mean but picture this scenario:
- You want to release a new version of Online Boutique
- You were working on a new feature / bug fix / whatever on a separate branch
- You forgot to change branch (but you have a few outstanding files not tracked by git yet)
- You run the releasing script that does some "magic" by checking out the `main` branch
- Uh oh... The files you were working on have carried over to the `main` branch (well, in fact they're not in _any_ branch still they're not tracked)
- Boom. You just released Online Boutique with unintended features / bug fixes / whatever :)

On another hand, the script could continue doing magic by running say `git add *` and `git stash` before the `git checkout` but still, I think that might create a lot of confusion (from the "user" which had no idea they were on the wrong branch).

I'll take an extra 10 seconds to manually change branch vs. potentially releasing something unintended and/or losing "in progress" files from an unrelated feature.

At minimum, there should be a hard-fail if there are _any_ uncommited changes:
```
# ensure there are no uncommitted changes
if [[ $(git status -s | wc -l) -gt 0 ]]; then
    echo "error: can't have uncommitted changes"
    exit 1
fi
```

mathieu-benoit: Good point, just added this check to make sure about that. Thansk!

NimJay (approved):

PR #1208, opened 2022-10-24
Simplify the grpc health probes kustomize overlay

mathieu-benoit:
Simplify the grpc health probes kustomize overlay, less actions by end users, because `tagSuffix` is not working upstream with Kustomize, we still need to update the Online Boutique version...


bourgeoisor (approved):
This LGTM!

NimJay (approved):

PR #1228, opened 2022-11-03
feat: Add skaffold kustomize

mmcloud:
### Background 

skaffold was not using the kustomize code. This PR gives an example of how
it can be used in skaffold.

Documentation on network policy was also updated to support local development with minikube.

### Change Summary

The PR will now apply network policies and test network policies

PR #1232, opened 2022-11-03
Use kustomize for skaffold local development

arbrown:
Adds kustomization.yaml to kubernetes-manifests folder Changes render target in skaffold.yaml to kustomize

### Background 
This change allows use of skaffold for local development with a different kustomize target than the default one in kustomize/

This allows us to keep separate "dev" manifests in kubernetes-manifests/ with clean "release" versions occasionally generated and stored in kustomize/base.  That way, when you deploy from kustomize/ you will not get any unrelease development noise/features.

### Additional Notes
This might not be an ideal solution, but I believe it is a step in the right direction.  Note that kustomize components are all in kustomize/components; there are no separate development and release versions.  Perhaps the modules could be moved to a dev folder and copied to kustomize/ on release, or perhaps that is less of a concern than ensuring that release manifests always point to pre-built known images.  I'm not sure.

### Related PRs or Issues 

#1228 is definitely related.  Using skaffold profiles might provide a better solution, but I'm not sure that it addresses the release/dev manifest separation discussed in **Background**.

### DIFF
--- a/kubernetes-manifests/kustomization.yaml
+++ b/kubernetes-manifests/kustomization.yaml
@@ -1,0 +1,40 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+ - adservice.yaml
+ - cartservice.yaml
+ - checkoutservice.yaml
+ - currencyservice.yaml
+ - emailservice.yaml
+ - frontend.yaml
+ - loadgenerator.yaml
+ - paymentservice.yaml
+ - productcatalogservice.yaml
+ - recommendationservice.yaml
+ - redis.yaml
+ - shippingservice.yaml
+components:
+# - ../kustomize/components/cymbal-branding
+# - ../kustomize/components/google-cloud-operations
+# - ../kustomize/components/memorystore
+# - ../kustomize/components/network-policies
+# - ../kustomize/components/service-accounts
+# - ../kustomize/components/spanner
+# - ../kustomize/components/container-images-tag
+# - ../kustomize/components/container-images-tag-suffix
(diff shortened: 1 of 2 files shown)


kubernetes-manifests/kustomization.yaml
```diff
@@ -0,0 +1,40 @@
...
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+ - adservice.yaml
+ - cartservice.yaml
+ - checkoutservice.yaml
+ - currencyservice.yaml
+ - emailservice.yaml
+ - frontend.yaml
+ - loadgenerator.yaml
```
NimJay, line 24:
```suggestion
# - loadgenerator.yaml # During development, the loadgenerator module inside skaffold.yaml will be used.
```

NimJay: **Context**
In https://github.com/GoogleCloudPlatform/microservices-demo/pull/609, we split the `skaffold.yaml` file into two modules: `app` and `loadgenerator`.
This means, today, we could run:
1. `skaffold run --default-repo=gcr.io/my-project --module=app` to develop without loadgenerator
2. `skaffold run --default-repo=gcr.io/my-project --module=app` to develop _with_ loadgenerator 
3. `skaffold run --default-repo=gcr.io/my-project` to develop _with_ loadgenerator  (identical to 2.)

The purpose of doing this was purely to showcase skaffold's "module" (or "multi-config") feature.

CC: @deesejohn (in case you had thoughts) :)

NimJay: **Thought:** 
We could also split the `loadgenerator` into its own module.
But we're have to make sure to include it by default inside [/kustomize/kustomization.yaml](https://github.com/GoogleCloudPlatform/microservices-demo/blob/main/kustomize/kustomization.yaml) because some tutorials/demo using Online Boutique prefer to have the `loadgenerator` create "traffic".

deesejohn: I think leaving the `loadgenerator` module as `rawYaml` in the [skaffold.yaml](https://github.com/GoogleCloudPlatform/microservices-demo/blob/main/skaffold.yaml) is a useful example for showing off both renders (`rawYaml` and `kustomize`). I don't see any harm in splitting it out into its own component though.

NimJay (approved):
Thanks for adding in my suggestion.
Looks good.
Approved!

NimJay (approved):
Thanks for adding in my suggestion.
Looks good.
Approved!

kubernetes-manifests/kustomization.yaml
NimJay: Thanks, @deesejohn, for your input. That helps (i.e., knowing that you're open to using either `rawYaml` or `kustomize` for the `loadgenerator`).

PR #1227, opened 2022-11-02
fix: Upgrade to skaffold/v3

mmcloud:
### Background 

Skaffold has released a major new update.
### Change Summary

Run `skaffold fix` with new skaffold v2.0.0 binarys. 

### Additional Notes

For upgrade notes see:
https://skaffold.dev/docs/upgrading/

PR #1248, opened 2022-11-08
feat: Add istio deployment

mmcloud:
### Background 

Istio was not being installed. The commit installs istio through helm charts using skaffold. 

### Change Summary

The root skaffold file points to the istio-manifests directory where  a second skaffold exists with the istio profile to istio


NimJay:
@mmcloud,
Again, thank you for both #1227 and #1228.
And thank you for implementing this new pull-request for Istio!

Unfortunately, I will have to close this pull-request since this is a large change that isn't fully aligned with our team's roadmap for Online Boutique. So sorry, Matthew!

**Some immediately thoughts:**
- We want to eventually get rids of [/istio-manifests](https://github.com/GoogleCloudPlatform/microservices-demo/tree/main/istio-manifests) and use a separate Kustomize component instead. We would just have to do an audit on which cloud.google.com tutorial (and other resources/demos/blogs/etc.) still use [/istio-manifests](https://github.com/GoogleCloudPlatform/microservices-demo/tree/main/istio-manifests).
- This pull-request is introducing Helm into this repo. We haven't properly assessed if this is the right decision for Online Boutique (e.g., how would Helm play with the [existing Kustomize Components](https://github.com/GoogleCloudPlatform/microservices-demo/tree/main/kustomize/components)? how should Helm best be used?).
- Helm requires users to install an additional tool. One large reason we introducing Kustomize earlier this year was because it's built into `kubectl`.

CC: @mathieu-benoit (in case you had thoughts. I know you were considering introducing Helm to our sample apps since Helm is widely used)

PR #1275, opened 2022-11-18
fix: Correct minikube default CNI

mmcloud:
the default is kindnet not flannel. ref: https://minikube.sigs.k8s.io/docs/handbook/network_policy/#:~:text=minikube%20allows%20users%20to%20create,at%20the%20end%20of%20development.

### Background 
Docs were incorrect


NimJay (approved):
According to the [official Kind documentation](https://kind.sigs.k8s.io/docs/user/configuration/#disable-default-cni), this is accurate — the default CNI is `kindnetd`.
Thank you, @mmcloud, for fixing this! :)

PR #1309, opened 2022-11-28
Clarify product-requirements.md

NimJay:
### Background 
* The `product-requirements.md` needs to clarify that the requirements are about the _default_ deployment of Online Boutique.
* This was inspired by a conversation we had with @boredabdel.

### Change Summary
* Clarify that Online Boutique welcomes changes as long as they are added as an "opt-in" feature — and is not builts into the default configuration of Online Boutique.

### Testing Procedure
* Read `product-requirements.md` as though you are a non-Googler that would like to contribute.

### DIFF
--- a/docs/product-requirements.md
+++ b/docs/product-requirements.md
@@ -8,7 +8,7 @@ Every change must:
 
 These requirements are about the default deployment (default configuration) of Online Boutique.
 Changes that will violate any of these rules should not be built into the default configuration of Online Boutique.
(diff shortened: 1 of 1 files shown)


NimJay:
Thanks for the quick review. And great suggestion, @bourgeoisor.
I've clarified based on your suggestion:
> ... if they align with the [purpose of Online Boutique](https://github.com/GoogleCloudPlatform/microservices-demo/blob/f200e4797a1cfa8e2be4088e009ca86a8dd763a7/docs/purpose.md).

PR #1310, opened 2022-11-28
Renovate for container images in Kubernetes manifests

mathieu-benoit:
https://docs.renovatebot.com/modules/manager/kubernetes/

As soon as it's merged in `main`, the goal is to catch any update of any container images referenced in any Kubernetes manifests. Like for example https://github.com/GoogleCloudPlatform/microservices-demo/blob/main/kustomize/components/google-cloud-operations/otel-collector.yaml#L71

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -39,5 +39,8 @@
   },
   "npm":{
     "addLabels": ["lang: nodejs"]
+  },
+  "kubernetes": {
+    "fileMatch”: [“(^|/)[^/]*\\.yaml$”] }
   }
 }
--- a/renovate.json
+++ b/renovate.json
@@ -41,6 +41,10 @@
     "addLabels": ["lang: nodejs"]
   },
   "kubernetes": {
-    "fileMatch": ["\\.yaml$"]
+    "fileMatch": ["\\.yaml$"],
+    "ignorePaths": [
+      "^release/**",
+      "^kustomize/base/**"
+    ]
   }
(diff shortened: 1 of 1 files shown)


mathieu-benoit:
Ready for review, thanks!

renovate.json
```diff
@@ -39,5 +39,8 @@
   },
   "npm":{
     "addLabels": ["lang: nodejs"]
+  },
+  "kubernetes": {
+    "fileMatch": ["\\.yaml$"]
```
NimJay, line 44: **Suggestion:**
We should ignore `release/` and `kustomize/base/` since these folders are only modified when we create new released.

NimJay: We can use [`ignorePath`](https://docs.renovatebot.com/configuration-options/#ignorepaths).

mathieu-benoit: Good point, just updated it accordingly.

NimJay (approved):
Looks great. Thanks for doing this, Mathieu.

mathieu-benoit:
FYI, like expected this update has been caught thanks to this PR: https://github.com/GoogleCloudPlatform/microservices-demo/pull/1313

mathieu-benoit:
We could also see that the Dependency Dashboard now has `Kubernetes` as `Detected dependencies`: https://github.com/GoogleCloudPlatform/microservices-demo/issues/728


image

PR #1312, opened 2022-11-29
Use `digest` instead of `tag` for container base images

mathieu-benoit:
Use `digest` instead of `tag` for container base images.

This follows best practices: https://cloud.google.com/architecture/using-container-images.

> We recommend that you pin your Docker images to an exact digest. By pinning to a digest you make your Docker builds immutable, every time you do a pull you get the same content.

And this will allow to catch as early as possible container base images updates when we target images such as: `eclipse-temurin:18`, etc. which are mutable. See illustration there too: https://github.com/GoogleCloudPlatform/microservices-demo/pull/1293#issuecomment-1324140441

Renovate will still update the base images for us: https://docs.renovatebot.com/docker/#digest-updating


mathieu-benoit:
Ready for review, thanks!

NimJay (approved):
You've covered all base images. All checks are passing. Staging URL works.
Thanks, Mathieu!
This is awesome. Approved.

PR #1281, opened 2022-11-21
Remove Google Cloud Debugger

NimJay:
## Background
* This fixes https://github.com/GoogleCloudPlatform/microservices-demo/issues/836.

## Summary of Changes
* I have removed all references to Cloud Debugger in documentation.
* I have removed all Cloud Debugger dependencies (only 3 microservices).
* I have removed all instances of `DISABLE_DEBUGGER`.
* I have regenerated the relevant `package-lock.json` files and `requirements.txt` file.
* I have removed the enablement of `clouddebugger.googleapis.com` inside `/terraform/`.
* I have **not** touched the `/kustomize/base/` and `/release/` folders.

## How to test/review
* Make sure the staging URL works fine.
* Review file changes.

## Additional Info
* After merging, we should close GitHub issues related to Cloud Debugger (see https://github.com/GoogleCloudPlatform/microservices-demo/issues/836 for more info).

PR #1316, opened 2022-11-29
Python 3.10

mathieu-benoit:
Python 3.10 for `emailservice`, `loadgenerator` and `recommendationservice`.

Thanks to https://github.com/GoogleCloudPlatform/microservices-demo/pull/1281 we could move forward with Python 3.10.

Note: same number of vulnerabilities still: 47 + the size of the container images increased a little bit with +1.6MB.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -10,7 +10,7 @@
     }
   },
   "constraints": {
-    "python": "~=3.9.0"
+    "python": "~=3.10.0"
   },
   "prConcurrentLimit":8,
   "stabilityDays":7,
--- a/src/emailservice/Dockerfile
+++ b/src/emailservice/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM python:3.9.15-slim as base
+FROM python:3.10.8-slim as base
 
 FROM base as builder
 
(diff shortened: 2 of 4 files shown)


renovate.json
```diff
@@ -10,7 +10,7 @@
     }
   },
   "constraints": {
-    "python": "~=3.9.0"
+    "python": "~=3.10.0"
```
NimJay, line 13: Praise: Good catch!

src/emailservice/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM python:3.9.15-slim as base
+FROM python:3.10.8-slim as base
```
NimJay, line 15: **Question:** If we merge this without the hash (base image digest), do you know if Renovate will include the digest when it updates it?

mathieu-benoit: Ah that's a good question, I don't know. I just resolved the merged by integrating the digest. Do you want we try with 1 of them, then merge this in `main` and see what will happen? I'm curious to know too, but the only way to see this in action is to merge in `main`. I could open a dedicated PR later if Renovate didn't do it for us to add the digest.

NimJay: Sweet, thanks for including the hashes/digests.

> Do you want we try with 1 of them (without the digest)
I personally don't have a yearning to know since all images now use digests.
But if _you're_ curious, feel free to create a PR. I will approve. :)

NimJay (approved):
Approved. Staging URL works!

Issue #1165, opened 2022-10-13
make `recommendation` service optional like `ad`


ameer00:
We would like to showcase that in microservices, not all services are created equal. For example services like ad and recommentaion can be run on cheaper cloud infra (for cost savings for example) and even if/when they go down, the rest of the application presses on without issues.

We would like to highlight various architecture patterns where some applications are critical and require highly reliable (multi cluster, multi regional etc) architectures vs others that are non-critical, can be put on cheaper infra (single cluster or single region e.g).

This is not a critical ask 😄  but would be nice to have.

PR #1332, opened 2022-12-01
make recommendation service optional

katzio:
### Background 
make recommendation service optional like ad

### Change Summary
Frontend template already supports missing recommendations, removed `renderHTTPError` but kept logs

### Additional Notes
should I add (error) log to [placeOrderHandler](https://github.com/GoogleCloudPlatform/microservices-demo/blob/main/src/frontend/handlers.go#L362)?

### Testing Procedure
Manually: turn off recommendations pod

### Related PRs or Issues 
addressing https://github.com/GoogleCloudPlatform/microservices-demo/issues/1165


mathieu-benoit:
Thanks for taking to initiative and opportunity to fix this, @katzio. Much appreciated!

mathieu-benoit (approved):
I haven't tested myself, but this LGTM!

We want the same behavior as `adservice`, and I can see that you did the exact same code like here https://github.com/GoogleCloudPlatform/microservices-demo/blob/main/src/frontend/handlers.go#L432.

Thanks again!

Before merging, let's have a final review by @NimJay on this.

mathieu-benoit:
Actually, I just tested in our test environment (for this PR), and we could see that there is no error anymore on the Product page is `recommendationservice` is scaled to 0 pods, see screenshot.
image

NimJay (approved):
@katzio, thanks so much for looking into this and implementing it.
I didn't think the change would be this simple.
This is great!

@mathieu-benoit, thank you for taking down the `recommendationservice` of the staging deployment. It worked!

Approved. Merging!

PR #1367, opened 2022-12-07
fix: Deploystack test enable test proejcts

tpryan:
### Background 
This allows the DeployStack test rig to run concurrent versions of the test.

Currently there is one test project per DeployStack project. In most cases this works just fine. However the microservices-demo often triggers multiple run of the test.  Since the opening step of the test is to clean and delete the resources this process creates, starting a second test while the first one is running will cause the first test to fail. This changes the process to dynamically generate/delete projects to get around this problem.

### Testing Procedure
Running the test should be the proper rest procedure.


.deploystack/test
```diff
@@ -78,10 +78,33 @@ function evalTest() {
...
+        echo
+    )
+    local __DATELABEL=$(date +%F)
+    local VALUE=ds-test-$__STACKSUFFIX-$__RANDOMSUFFIX
+    local __BA=$(gcloud beta billing accounts list --format="value(ACCOUNT_ID)" --limit=1 | xargs)
+   
+    gcloud projects create $VALUE --labels="deploystack-disposable-test-project=$__DATELABEL" --folder="155265971980"
+    gcloud beta billing projects link $VALUE --billing-account=$__BA
+    eval $__resultvar="'$VALUE'"
+}
```
NimJay, line 95: **Praise:** Thanks for modularizing this code into a separate function. It made it easier for me to review the code.

.deploystack/test.yaml
```diff
@@ -16,9 +16,6 @@
 # in the testing rig
 
 steps:
-   - name: 'gcr.io/cloudshell-images/cloudshell:latest'
-     entrypoint: bash
-     args: [ '.deploystack/clean' ]
```
NimJay, line 21: **Thought:**
I wonder if it makes more sense to have these tests live in a separate DeployStack-owned repository.
That `git` repo would include all similar tests for DeployStack-enabled applications.
The pros would be:
1. I (and other sample app owners) won't be a bottleneck to your team's changes to these tests.
2. I assume there's some common code between each of the test across the apps. Modularizing and reusing that code would be easier (since all the code would live in the same repo).

This is just an idea. No action needed. :)

.deploystack/test
```diff
@@ -112,20 +141,33 @@ section_open "Testing Online Boutique's services are running"
...
 section_open "Testing Google Kubernetes Engine cluster does NOT exist"
     evalTest 'gcloud container clusters describe online-boutique --format="value(name)" --region $REGION' "EXPECTERROR"
 section_close
 
+# This is only needed if you tests fail alot because of overlaping runs of the
+# same set of tests. Really don't do this if you don't want to severly irritate 
+# @tpryan
+section_open "Delete Test Project"
+    gcloud projects delete $PROJECT -q
+section_close
```
NimJay, line 169: **Thought:**
Previously, we had a `clean` bash script that cleaned up GCP resources.
The `clean` bash script ran as a separate step inside the Google Cloud Build Trigger (`.deploystack/test.yaml`).
Also, `clean` ran _before_ `test`, so I'm guessing failures in `test` didn't step `clean` from running.
It it possible that we end up with if failures occur in the `test` script before we reach `gcloud projects delete $PROJECT -q`?
```
gcloud container clusters delete online-boutique --region $REGION -q
```

Just a thought. No action needed.

NimJay (approved):
Thank you so much, @tpryan, for addressing this issue!

Approved. But if possible, could you elaborate in the pull-request description (e.g., include a "Background" section with some more info on what's happening)?
It would help others on my team understand what's this pull-request is about.

All my other comments can be ignored. :)

.deploystack/test.yaml
tpryan: I've done that with the core functionality. But I've never figured out a good solution for sharing bash code. Any ideas you have would be super welcome by me.

.deploystack/test
tpryan: Yes that's actually a feature, not a bug.  If a test fails the project remains for forensics.

The DeployStack platform will prune these projects after some predetermined length of time. I need to write up the procedure and document how to get access to the projects and pause deletion. .

NimJay: > "If a test fails the project remains for forensics."

That totally makes sense!

PR #1409, opened 2022-12-15
chore: update config for metrics purposes

tpryan:
### Background 
Cannot record metric impact of running this through Deploystack - no way to count api usage without this string in the config file

### Change Summary
Added a line to the deploystack config to make sure that these calls are labeled properly in our systems.  This is a super minimal fix that only affects how this stack's deployments are metricked.

### DIFF
--- a/.deploystack/deploystack.yaml
+++ b/.deploystack/deploystack.yaml
@@ -1,6 +1,7 @@
 # The fields inside this deploystack.yaml file are documented in https://github.com/GoogleCloudPlatform/deploystack.
 
 title: Microservices Demo (Online Boutique)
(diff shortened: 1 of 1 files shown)


NimJay (approved):
Thanks, @tpryan, for continually supporting the "Online Boutique via DeployStack" user journey!

Approved.

This sounds like a very DeployStack-specific change, so I don't expect Online Boutique maintainers (including myself) to understand the details. I'll share [this old comment of mine](https://github.com/GoogleCloudPlatform/microservices-demo/pull/1367#discussion_r1042740521) for any Online Boutique maintainer that might have the same idea.

PR #1427, opened 2022-12-22
Fix `/hack/make-helm-chart.sh`

NimJay:
### Background 
* `make-helm-chart.sh` is run when a maintainer creates a new release of Online Boutique.
* `make-helm-chart.sh` bumps up the version of the Helm Chart published at [`oci://us-docker.pkg.dev/online-boutique-ci/charts/onlineboutique`](https://pantheon.corp.google.com/artifacts/docker/online-boutique-ci/us/charts/onlineboutique).
* This pull-requests makes minor fixes to `make-helm-chart.sh`.

### Change Summary
* See the comments I've made in the file changes.

### Testing Procedure
* No testing from reviewer required.
* These changes were tested when I did the previous release of Online Boutique.

### DIFF
--- a/hack/make-helm-chart.sh
+++ b/hack/make-helm-chart.sh
@@ -25,9 +25,9 @@ TAG="${TAG:?TAG env variable must be specified}"
 HELM_CHART_REPO="us-docker.pkg.dev/online-boutique-ci/charts"
 
 cd helm-chart
-sed -i "s/^appVersion:.*/appVersion: \"${TAG}\"/" Chart.yaml
-sed -i "s/^version:.*/version: ${TAG:1}/" Chart.yaml
+gsed -i "s/^appVersion:.*/appVersion: \"${TAG}\"/" Chart.yaml
+gsed -i "s/^version:.*/version: ${TAG:1}/" Chart.yaml
 helm package .
-helm push onlineboutique-$TAG.tgz oci://$HELM_CHART_REPO
+helm push onlineboutique-${TAG:1}.tgz oci://$HELM_CHART_REPO
 
 log "Successfully built and pushed the Helm chart."


hack/make-helm-chart.sh
```diff
@@ -25,9 +25,9 @@ TAG="${TAG:?TAG env variable must be specified}"
 HELM_CHART_REPO="us-docker.pkg.dev/online-boutique-ci/charts"
 
 cd helm-chart
-sed -i "s/^appVersion:.*/appVersion: \"${TAG}\"/" Chart.yaml
-sed -i "s/^version:.*/version: ${TAG:1}/" Chart.yaml
+gsed -i "s/^appVersion:.*/appVersion: \"${TAG}\"/" Chart.yaml
+gsed -i "s/^version:.*/version: ${TAG:1}/" Chart.yaml
```
NimJay, line 29: We `hack` script [uses `gsed`](https://github.com/GoogleCloudPlatform/microservices-demo/tree/main/hack#hack) instead of `sed`.
Using `gsed` allows us to run the script `sed` commands on both Linux _and_ Mac.

hack/make-helm-chart.sh
```diff
@@ -25,9 +25,9 @@ TAG="${TAG:?TAG env variable must be specified}"
...
 helm package .
-helm push onlineboutique-$TAG.tgz oci://$HELM_CHART_REPO
+helm push onlineboutique-${TAG:1}.tgz oci://$HELM_CHART_REPO
```
NimJay, line 31: The `TAG` environment variable's value looks like: `v0.5.1`.
Using `${TAG:1}` means that we remove the `v`.

mathieu-benoit (approved):

hack/make-helm-chart.sh
mathieu-benoit: Good catch!

NimJay:
Thanks for the quick review/approva, @mathieu-benoit! :)

Issue #1444, opened 2023-01-09
OTel Collector lacks permissions on GKE


arbrown:
### Describe the bug 
The OpenTelemetry Colector fails to export spans to cloud trace on GKE Autopilot clusters.  It fails with the error:

```
failed to export to Google Cloud Trace: rpc error: code = PermissionDenied desc = The caller does not have permission"
```

This worked previously, and seems to be related to a change in Workload Identity permissions on GKE autopilot.  Note that trace exports work as intended on GKE standard.

### To Reproduce 
Deploy Online Boutique with the googlecloudoperations kustomize module on a GKE autopilot cluster.  Note the lack of spans in Cloud Trace, and observe the error messages in the logs for the collector.

NimJay:
(We talked about this offline, but I'll mention it here for visibility.)

You mentioned:
> "Note that trace exports work as intended on GKE standard."

I was able to reproduce this permission issue with a **standard** GKE cluster — not autopilot.

### Reproduction Details

GKE version: 1.24.7-gke.900	
Num of nodes: 2
Branch/commit used: `main`
Error message in `opentelemetrycollector`:
```
2023/01/09 21:09:51 failed to export to Google Cloud Trace: rpc error: code = PermissionDenied desc = The caller does not have permission
```

NimJay:
I should clarify: the standard GKE cluster I mentioned in my previous comment has Workload Identity enabled.

<img width="390" alt="Screenshot 2023-01-11 at 9 48 20 AM" src="https://user-images.githubusercontent.com/10292865/211836592-f52642a1-8559-4c6e-b8d2-5378d99fbae9.png">

christopherhendrich:
I am seeing the same error with standard clusters that have WI enabled. 
otel-collector uses the default KSA of the namespace, correct?
I tried to run the below without any success. The GSA has the permissions to cloud trace as described as in the readme. 
```
gcloud iam service-accounts add-iam-policy-binding ${GSA_NAME}@${PROJECT_ID}.iam.gserviceaccount.com \
    --member "serviceAccount:${PROJECT_ID}.svc.id.goog[${ONLINEBOUTIQUE_NAMESPACE}/default]" \
    --role roles/iam.workloadIdentityUser
    ```

**Comment by arbrown** [comment]
Hi Christopher,

I think you're on the right track.  As long as all the variables in your command are properly set, the only other thing to do is update the Kubernetes service account with a similar command:
```
[...]

PR #1461, opened 2023-01-23
Add documentation for Workload Identity fix

arbrown:
### Background 
Previously, workloads would use the default compute service account on GKE, which allowed them to use Google APIs by default.  This changed for some versions of GKE Autopilot (and will change for Standard) and requires some extra work to allow the OTel collector to send traces to Cloud Trace.

### Related PRs or Issues 
#1444 

### Future Work
I think this is just a proper fix until we get a better idea about how to handle these Workload Identity changes.  Using the default service account is likely not an appropriate long-term solution, but works for now until we have a unified approach for handling service accounts and permissions in the project.

### DIFF
--- a/kustomize/components/google-cloud-operations/README.md
+++ b/kustomize/components/google-cloud-operations/README.md
@@ -84,3 +84,27 @@ Currently, this component adds a single collector service which collects traces 
 ![Collector Architecture Diagram](collector-model.png)
 
 If you wish to experiment with different backends, you can modify the appropriate lines in [otel-collector.yaml](otel-collector.yaml) to export traces or metrics to a different backend.  See the [OpenTelemetry docs](https://opentelemetry.io/docs/collector/configuration/) for more details.
+
+## Workload Identity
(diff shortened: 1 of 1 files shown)


NimJay (approved):
Thank you for this solution, @arbrown (and for acknowledging that this is a temporary solution). I have made minor edits to the README.md in https://github.com/GoogleCloudPlatform/microservices-demo/pull/1461/commits/498300b94471c0353e93a87864a0ac386f7502ee.

### Manually Tested

I have tested the commands on my own GKE cluster (standard with workload identity enabled).
Tracing worked! 👍 

Screenshot 2023-01-24 at 10 52 31 AM

### Log Severity

All the logs output by by the `opentelemetrycollector` Pod have a severity of `"ERROR"` — even if the log itself is not an `ERROR` (e.g., in the screenshot below, I've expanded an `INFO` log).
Have you been experiencing the same?
If so, I can create a GitHub issue so we can track this issue.

Screenshot 2023-01-24 at 10 56 55 AM

PR #1574, opened 2023-03-02
fix: rename deprecated script names

tpryan:
### Background 
DeployStack is deprecating some names of its scripts. They were redundant. (preinitscript.sh)   The new name (preinit.sh) already works and works the same as what it is replacing.


NimJay (approved):
The "script" in "preinitscript.sh" _does_ feel redundant. 
Thanks for addressing this. Approved. Merging.

PR #1693, opened 2023-04-17
Fix Hero Logo: for dark background.

Deleplace:
GitHub has provided dark theme for users like me who prefer dark, which is great and sometimes comes with challenges for existing images.

The current logo is dark red, which (on some devices) makes it difficult to see at all.

### Dark

image

### Light

image

## Suggested fix

Still 1 transparent SVG file, that looks okay in both light theme and dark theme: 

### Dark

image

### Light

image


bourgeoisor:
Thank you for your contribution. I think it would be much more readable if we used the original logo, but with a 2px white outline throughout. That outline would be invisible in light mode, and help readability with dark mode.

Is that something we can try?

Deleplace:
Applied white stroke as suggested

image

bourgeoisor:
Looks better-- @NimJay what do you think?

NimJay (approved):
Thanks for addressing this @Deleplace!
Looks great!

Good suggestion, @bourgeoisor!

Approved.

PR #1720, opened 2023-04-21
Rename . gitattributes to .gitattributes

bourgeoisor:
This PR fixes a typo in the name of the gitattributes file (it had a stray whitespace between the dot and the filename).


NimJay (approved):
How did you even notice that? Nice!

Issue #1772, opened 2023-05-17
Implement `FRONTEND_MESSAGE` environment variable


NimJay:
### Describe request or inquiry 
* See Google-internal change: [cl/529866615](http://cl/529866615)
* As part of the tutorial in [cl/529866615](http://cl/529866615), we want the user to make a change to the Kubernetes YAML files that would result in the rolling update of a microservice.
* We (@villasenor and I) propose the following new user experience:
    1. the user deploys Online Boutique (The aforementioned tutorial will use the variation of Online Boutique found in [Jump Start Solution, _Ecommerce web app deployed on Kubernetes_](https://github.com/GoogleCloudPlatform/terraform-ecommerce-microservices-on-gke).)
    1. the user edits a YAML file to set an environment variable called `FRONTEND_MESSAGE` in the `frontend` Deployment (see code snippet below)
    1. the user will `kubectl apply -f` the newly edited YAML file
    1. the user sees the value of `FRONTEND_MESSAGE` output on the frontend of Online Boutique
* Code snippet:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: frontend
spec:
  ...
  template:
    ...
    spec:
      ...
      containers:
          ...
          env:
            ...
          - name: FRONTEND_MESSAGE
            value: "The user-provided message that will be displayed on the frontend."
```

### User interface

* The message will be displayed at the very top of every page. See screenshot below:

<img src="https://github.com/GoogleCloudPlatform/microservices-demo/assets/10292865/c3a69639-b493-4575-815c-5c16940c5744" height="300px" alt="Screenshot of Online Boutique's homepage where the very top of the page has a maroon banner with a message.">
 
* We are reusing the UI from the old "Free Shipping" banner that we removed in [pull-request 800](https://github.com/GoogleCloudPlatform/microservices-demo/pull/800).

PR #1773, opened 2023-05-17
Implement `FRONTEND_MESSAGE` env var

NimJay:
* This fixes https://github.com/GoogleCloudPlatform/microservices-demo/issues/1772.
* I have not created a separate Kustomize component for this since a dedicated Kustomize component would be overkill.

### How to review
* You don't need to test in your own cluster.
* I have deployed Online Boutique using the `FRONTEND_MESSAGE` feature at the [IP address found in this gPaste](http://gpaste/5955576767971328).

### DIFF
--- a/kubernetes-manifests/frontend.yaml
+++ b/kubernetes-manifests/frontend.yaml
@@ -79,13 +79,15 @@ spec:
           - name: AD_SERVICE_ADDR
             value: "adservice:9555"
           # # ENV_PLATFORM: One of: local, gcp, aws, azure, onprem, alibaba
-          # # When not set, defaults to "local" unless running in GKE, otherwies auto-sets to gcp 
-          # - name: ENV_PLATFORM 
+          # # When not set, defaults to "local" unless running in GKE, otherwies auto-sets to gcp
+          # - name: ENV_PLATFORM
           #   value: "aws"
           - name: ENABLE_PROFILER
             value: "0"
           # - name: CYMBAL_BRANDING
           #   value: "true"
+          # - name: FRONTEND_MESSAGE
+          #   value: "Replace this with a message you want to display on all pages."
           resources:
             requests:
               cpu: 100m
--- a/src/frontend/handlers.go
+++ b/src/frontend/handlers.go
@@ -40,8 +40,9 @@ type platformDetails struct {
 }
 
 var (
-	isCymbalBrand = "true" == strings.ToLower(os.Getenv("CYMBAL_BRANDING"))
-	templates     = template.Must(template.New("").
+	frontendMessage = strings.TrimSpace(os.Getenv("FRONTEND_MESSAGE"))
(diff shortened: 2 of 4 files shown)


kubernetes-manifests/frontend.yaml
```diff
@@ -79,13 +79,15 @@ spec:
...
           #   value: "aws"
           - name: ENABLE_PROFILER
             value: "0"
           # - name: CYMBAL_BRANDING
           #   value: "true"
+          # - name: FRONTEND_MESSAGE
+          #   value: "Replace this with a message you want to display on all pages."
           resources:
             requests:
               cpu: 100m
```
NimJay, line 93: Reminder: change to this `frontend.yaml` will eventually (automatically) trickle down into `/kustomize/base/frontend.yaml` and `/release/kubernetes-manifests.yaml` when we create the next release of Online Boutique.

src/frontend/handlers.go
```diff
@@ -40,8 +40,9 @@ type platformDetails struct {
 }
 
 var (
-	isCymbalBrand = "true" == strings.ToLower(os.Getenv("CYMBAL_BRANDING"))
-	templates     = template.Must(template.New("").
+	frontendMessage = strings.TrimSpace(os.Getenv("FRONTEND_MESSAGE"))
+	isCymbalBrand   = "true" == strings.ToLower(os.Getenv("CYMBAL_BRANDING"))
+	templates       = template.Must(template.New("").
```
NimJay, line 45: The lines starting in:
```
isCymbalBrand ...
template ...
```
didn't change logically. The "change" we're seeing = spacing/formatting changes.

src/frontend/templates/header.html
```diff
@@ -46,6 +46,13 @@
 
 <body>
     <header>
+        {{ if $.frontendMessage }}
+        <div class="navbar">
+            <div class="container d-flex justify-content-center">
+                <div class="h-free-shipping">{{ $.frontendMessage }}</div>
+            </div>
+        </div>
+        {{ end }}
```
NimJay, line 55: We don’t need to sanitize `frontendMessage`. HTML injection is prevented by the [Golang html/template](https://pkg.go.dev/html/template).

villasenor (commented):
LGTM! Looks awesome.

src/frontend/templates/header.html
NimJay, line 55: A note about `{{ if $.frontendMessage }}`: we `$.frontendMessage` will evaluate to false if the string is empty, [according to the docs](https://pkg.go.dev/text/template#hdr-Actions) (and my manual testing):
```
{{if pipeline}} T1 {{end}}
	If the value of the pipeline is empty, no output is generated;
	otherwise, T1 is executed. The empty values are false, 0, any
	nil pointer or interface value, and any array, slice, map, or
	string of length zero.
	Dot is unaffected.
```

bourgeoisor (approved):
This LGTM!

PR #1688, opened 2023-04-12
Add .gitattributes to convert "CR + LF" to "LF"

NimJay:
### Background 
* See my [GitHub comment here](https://github.com/GoogleCloudPlatform/microservices-demo/issues/1434#issuecomment-1377451183).
* LF = "Line Feed" character (moves the cursor to the next line)
* CR = "Carriage Return" character (brings the cursor to the start of the current line)
* For marking the end of a line:
    * Linux & Mac use LF.
    * Windows uses CRLF.
* For more context, see [this StackOverflow post](https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/).

### Fixes 

#1434

### Change Summary
* Add a `.gitattributes` file that converts "CR + LF" to "LF".

### Testing Procedure
* [This comment](https://github.com/GoogleCloudPlatform/microservices-demo/issues/1434#issuecomment-1377472080) confirms that this fix likely works.

### Related PRs or Issues 
* Old, similar issue: #196.

Issue #1347, opened 2022-12-02
./gradlew: not found local cluster deployment


anasj25:
Step 4/18 : COPY gradle gradle
 ---> ca2a614ce472
Step 5/18 : RUN chmod +x gradlew
 ---> Running in 4cfbc82c2da9
Removing intermediate container 4cfbc82c2da9
 ---> 4ab32a37dc28
Step 6/18 : RUN ./gradlew downloadRepos
 ---> Running in 374228cb6f66
/bin/sh: 1: ./gradlew: not found
The command '/bin/sh -c ./gradlew downloadRepos' returned a non-zero code: 127

mathieu-benoit:
Thanks @anasj25 for reporting this issue!

Just to confirm, could you clarify which command you are running please?

I just tested `cd src/adservice && docker build -t adservice .` and it's working successfully without the error/issue you are facing:
```
Step 4/18 : COPY gradle gradle
 ---> 6a53b84138b6
Step 5/18 : RUN chmod +x gradlew
 ---> Running in 7c0aa69d8b5c
Removing intermediate container 7c0aa69d8b5c
 ---> 5514d4121c79
Step 6/18 : RUN ./gradlew downloadRepos
 ---> Running in 78bbabedbb48
Downloading https://services.gradle.org/distributions/gradle-7.6-bin.zip
...........10%............20%...........30%............40%............50%...........60%............70%............80%...........90%............100%

Welcome to Gradle 7.6!
```

Also, could you confirm that you have the latest version of Online Boutique?

Thanks for providing more context and information to help us diagnosing this issue.

anasj25:
Hi Mathieu,
I am on windows and I ran the skaffold run command. And yes I cloned the latest Online Boutique repo on my PC.

NimJay:
Hey @anasj25,
* Sorry for the extremely late response!
* Thanks for providing additional info. :)
* The error message is similar to error message seen in [issue 1434](https://github.com/GoogleCloudPlatform/microservices-demo/issues/1434#issuecomment-1377451183).
```
/bin/sh: 1: ./gradlew: not found
```
* Could you retry (re-`git clone`) with the [following workaround](https://github.com/GoogleCloudPlatform/microservices-demo/issues/1434#issuecomment-1377451183)?
```
git clone https://github.com/GoogleCloudPlatform/microservices-demo.git --config core.autocrlf=true
```
* This issue seems related to the following discrepancy: 
[...]

PR #1806, opened 2023-05-29
Convert CRLF to LF

NimJay:
### Background 
* This fixes [issue 1347](https://github.com/GoogleCloudPlatform/microservices-demo/issues/1347) — which [pull-request 1688](https://github.com/GoogleCloudPlatform/microservices-demo/pull/1688) didn't fully fix.
* `LF` (invisible character) is used by **Mac/Linux** to denote the end of a line in a file.
* `CRLF` (invisible characters) is used by **Windows** to denote the end of a line in a file.
* What the `.gitattributes` file from [pull-request 1688](https://github.com/GoogleCloudPlatform/microservices-demo/pull/1688) does:
    * it ensures newly committed text files (i.e., not binaries/PNGs/JPEG) use `LF` in remote — not necessarily locally
    * it **does not** convert _existing_ files that use `CRLF` to `LF`
    * it also used `CRLF` locally on Windows and `LF` locally on Mac and Linux
* I used `git ls-files --eol` to check which text files (ignoring png/jpeg/jar/tgz/etc.) were still using `CRLF`. I found that the following files were still using `CRLF`:
    * `src/cartservice/cartservice.sln`
    * `src/adservice/gradlew.bat`
* @marcghanime mentioned [in this comment](https://github.com/GoogleCloudPlatform/microservices-demo/issues/1347#issuecomment-1562847984) that they were able to fix the issue by converting `src\adservice\gradlew.bat` (and 2 other files) to `LF` (from `CRLF`).

### Change Summary
 * I ran: `git add --renormalize .` to convert the following files to `LF` (from `CRLF`):
    * `src/cartservice/cartservice.sln`
    * `src/adservice/gradlew.bat`
* change the `.gitattributes` file such that even local files on Windows use `LF` — not `CRLF`.

### Testing Procedure
* We should ensure:
     * GitHub checks of this PR were able to successfully build the `adservice` Dockerfile 
     * we are able to `docker build` the `adservice` on Windows


bourgeoisor (approved):
LGTM, thank you for checking this!

NimJay:
Thank you, @bourgeoisor, for the quick review.
I've included another minor change (in the `.gitattributes` file that ensures Windows uses `LF` locally).
[This comment](https://github.com/GoogleCloudPlatform/microservices-demo/issues/1347#issuecomment-1570350794) has more info.
Merging.

PR #2013, opened 2023-08-09
Fix deprecated methods in productcatalog

bourgeoisor:
This PR fixes a few deprecated methods in the `productcatalog`:
- The gRPC `WithInsecure()` now takes credentials as argument
- `ReadFile()` should now be called form the `os` package

### DIFF
--- a/src/productcatalogservice/product_catalog.go
+++ b/src/productcatalogservice/product_catalog.go
@@ -62,7 +62,6 @@ func (p *productCatalog) GetProduct(ctx context.Context, req *pb.GetProductReque
 func (p *productCatalog) SearchProducts(ctx context.Context, req *pb.SearchProductsRequest) (*pb.SearchProductsResponse, error) {
 	time.Sleep(extraLatency)
 
-	// Interpret query as a substring match in name or description.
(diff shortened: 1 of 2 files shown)


src/productcatalogservice/product_catalog.go
```diff
@@ -62,7 +62,6 @@ func (p *productCatalog) GetProduct(ctx context.Context, req *pb.GetProductReque
 func (p *productCatalog) SearchProducts(ctx context.Context, req *pb.SearchProductsRequest) (*pb.SearchProductsResponse, error) {
 	time.Sleep(extraLatency)
 
-	// Interpret query as a substring match in name or description.
```
NimJay, line 65: **Question (just curious, no change needed):** Did you remove this comment because it's false or because it feels unnecessary?

NimJay (approved):
Thanks for this update! Looks good.
All checks are passing + staging URL works.

src/productcatalogservice/product_catalog.go
bourgeoisor: The 3 lines were already self-explanatory

PR #1844, opened 2023-06-15
Release v0.8.0

NimJay:
We're creating a new release: `v0.8.0`.

## Minor Version Bump 

We're bumping the minor version because of https://github.com/GoogleCloudPlatform/microservices-demo/pull/1837.

## How To Review This Pull-Request
* [x] Make sure [Draft release notes](https://github.com/GoogleCloudPlatform/microservices-demo/releases) look good. 
* [x] Make sure the files changes look good (e.g., include all **11** microservices).
* [x] Make sure all 11 microservices have `v0.8.0` Docker images in [our Container Registry](https://pantheon.corp.google.com/gcr/images/google-samples/global/microservices-demo?project=google-samples).

PR #2138, opened 2023-09-25
feat: renovate otel grouping and pip-compile fixes

muncus:
Moving to json5 allows comments in this config file.

OTel dependency groups avoid PRs with broken tests because the OTel packages
have tight version coupling.

Also updates the way language labels are added, as suggested by renovate's
"config migrator"


.github/renovate.json5
```diff
@@ -0,0 +1,66 @@
+{
+  "extends": [
+    "config:recommended"
+  ],
+  "pip-compile": {
+    "enabled": true,
+    "fileMatch": ["(^|/)requirements\\.in$"]
```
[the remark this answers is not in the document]
muncus: lock file maintenance is enabled by default for this manager:
https://docs.renovatebot.com/modules/manager/pip-compile/#default-config

so i don't *think* we need it, but if we do I'm happy to add it in a followup PR.

(and questions are always welcome, even on drafts!)

.github/renovate.json5
```diff
@@ -0,0 +1,66 @@
...
+  // Dependency Dashboard should be excluded from SLO calculation.
+  "dependencyDashboardLabels": [
+    "type: process",
+  ],
+  "packageRules": [
+    // Label PRs based on the languages they contain
+    {"matchCategories": ["python"], "addLabels": ["lang: python"]},
+    {"matchCategories": ["java"], "addLabels": ["lang: java"]},
+    {"matchCategories": ["golang"], "addLabels": ["lang: go"]},
+    {"matchCategories": ["docker"], "pinDigests": true},
```
NimJay, line 40: **Issue:** What about labels for `dotnet` and `nodejs`?

.github/renovate.json5
```diff
@@ -0,0 +1,66 @@
...
+     "labels":[
+       "type:security"
+     ],
+     "stabilityDays":0
+  },
+  "labels": ["dependencies"],
+  "kubernetes": {
+    "fileMatch": ["\\.yaml$"],
+    "ignorePaths": [
+      "kubernetes-manifests/**"
```
NimJay, line 28: **Issue:**
```suggestion
      "release/**",
      "kustomize/base/**"
```

* Most files inside `release/` folder and the `kustomize/base/` only ever get modified when we make new release of Online Boutique. 
* Example release PR: https://github.com/GoogleCloudPlatform/microservices-demo/pull/1844/files
* _[/docs/releasing/](https://github.com/GoogleCloudPlatform/microservices-demo/tree/main/docs/releasing)_ contains more info about what happens during releases. But the tl;dr: 
- we run a script called `make-release.sh`, which 
    - builds Online Boutique locally
    - pushes those built images to gcr.io/google-samples
    - [copies the Kubernetes manifests inside `/kubernetes-manifests/`](https://github.com/GoogleCloudPlatform/microservices-demo/blob/release/v0.8.0/docs/releasing/make-release-artifacts.sh#L97) into `release/` folder and the `kustomize/base/` .

NimJay: Update: according to [PR 1325](https://github.com/GoogleCloudPlatform/microservices-demo/pull/1325), the reason we ignored `kubernetes-manifests/**` is to avoid getting too many Renovate PRs. Sounds reasonable. 
But let's try configuring Renovate to work on `kubernetes-manifests/**`.

.github/renovate.json5
muncus: Good catch, thanks! I've added these now.

.github/renovate.json5
muncus: Nim, thanks for all the context!
I think there's more we can do to help with Renovate for releases, but i'll need to understand the process a bit more before moving forward.
I'll move this discussion into my renovate proposal, and we can discuss further there.

NimJay (approved):
Approved! Thanks for addressing my concerns.
But more importantly, thank you for working on these Renovate improvements, @muncus!
Merging.

PR #2168, opened 2023-10-03
`cartservice` - Memorystore (Redis) 7.0

mathieu-benoit:
Memorstore (Redis) now [supports Redis 7.0](https://cloud.google.com/blog/products/databases/redis-7-0-comes-to-memorystore/).

The in-cluster version of the `redis-cart` has been using Redis 7.0 by default for a while now because of using [`redis:alpine`](https://github.com/GoogleCloudPlatform/microservices-demo/blob/main/kubernetes-manifests/redis.yaml#L42) (i.e. `latest` being 7.0). So this is confirming that `cartservice` is working well with Redis 7.0, no issues.

I also tested this as part of [this "tutorial" (provisioning via `gcloud`)](https://medium.com/google-cloud/use-google-cloud-memorystore-redis-with-the-online-boutique-sample-on-gke-82f7879a900d?sk=8874a67a93cf617367e744708d34ad44). I also did the test of provisioning a Memorystore (Redis) instance via Terraform. All good!

### DIFF
--- a/kustomize/components/memorystore/README.md
+++ b/kustomize/components/memorystore/README.md
@@ -23,7 +23,7 @@ gcloud redis instances create redis-cart \
     --size=1 \
     --region=${REGION} \
     --zone=${ZONE} \
-    --redis-version=redis_6_x
+    --redis-version=redis_7_0
 ```
 
 _Note: You can also find in this repository the Terraform script to provision the Memorystore (Redis) instance alongside the GKE cluster, more information [here](/terraform)._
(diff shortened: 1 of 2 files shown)


mathieu-benoit:

NimJay (approved):
Thank you, @mathieu-benoit, for this update!
I was hoping to preserve the `_x` minor version (to prepare for when/if 7.1 is a thing), but it looks like `redis_7_x` isn't supported, based on [docs here](https://cloud.google.com/sdk/gcloud/reference/redis/instances/create#--redis-version:~:text=node%2Dmax%2Dentries.-,%2D%2Dredis%2Dversion,-%3DVERSION).

Looks great! Approved.

PR #2214, opened 2023-10-17
feat(renovate): move to per-language PR grouping, and weekly schedule.

muncus:
Renovate will group PRs by language, so we should receive one PR per week, per
language. 

"language" also includes docker, so digest pinning should all be grouped in a
single PR as well.


.github/renovate.json5
```diff
@@ -50,21 +52,51 @@
       "matchDepTypes": ["devDependencies"],
       "automerge": true
     },
-    // OpenTelemetry has tight version coupling.
-    // For go, js, and dotnet, there are 'monorepo:opentelemetry-$LANG' presets already included.
+    // OpenTelemetry
```
NimJay, line 55: **Suggestion:**
I actually liked the comments that you had.
Unfortunately, it won't be obvious to most maintainers of this repo (including me) that this is _**grouping**_ Renovate PRs — just from looking at the JSON. 
```suggestion
    // Group OpenTelemetry's tightly-coupled Python updates
```

muncus: The functionality here has changed, and this rule is actually *only* including pre-release versions of instrumentation packages.
The spirit of your comment is valid, so i've updated these to be descriptive comments, rather than "headers".

.github/renovate.json5
```diff
@@ -1,7 +1,11 @@
 {
   "extends": [
-    "config:recommended"
+    "config:recommended",
+    ":approveMajorUpdates",
```
NimJay, line 4: **1. Question:** I don't understand the [docs on `approveMajorUpdates`](https://docs.renovatebot.com/presets-default/#approvemajorupdates).
Does this mean we have to manually tick a checkbox inside https://github.com/GoogleCloudPlatform/microservices-demo/issues/728 to trigger major updates?

> Require dependency dashboard approval for `major` updates.

**2. Suggestion:** Are we able to exclude major versions from the 6 groupings you created instead? (so we automatically get major pull-requests, but via separate pull-requests)

.github/renovate.json5
NimJay: Thank you! That helps!

.github/renovate.json5
NimJay: Update: Ah, I see that the groupings already exclude `major` updates. Nice!

NimJay: That said, I think we should embrace the automatic `major` updates
```suggestion
```

NimJay (approved):
The only change I think we should make is regarding `approveMajorUpdates`. :)

Other than that, this looks awesome!
Thanks again so much, @muncus, for looking into our Renovate configs.
Your work really going to help our team.

.github/renovate.json5
```diff
@@ -1,7 +1,11 @@
...
   ],
+  // Run once a week, outside typical office hours.
+  // TZ is UTC unless specified.
+  "schedule": [ "after 10pm on Tuesday" ],
```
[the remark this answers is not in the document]
muncus: scheduling affects the groupings - only changes found in a single "schedule interval" are considered for grouping, so longer schedule periods will help us get better grouping.
We can run renovate against the repo off-schedule whenever we'd like.

Issue #674, opened 2021-12-29
Use `pino` instead of `console.log`


NimJay:
### Describe request or inquiry 

* Both JavaScript/Node.js microservices (`paymentservice` and `currencyservice`) use `console.log`.
* See files:
  * [`/src/paymentservice/index.js`](https://github.com/GoogleCloudPlatform/microservices-demo/blob/80acbf22ce074dce22e03dd38efbf156f88eae99/src/paymentservice/index.js#L21)
  * [`/src/currencyservice/server.js`](https://github.com/GoogleCloudPlatform/microservices-demo/blob/80acbf22ce074dce22e03dd38efbf156f88eae99/src/currencyservice/server.js#L18)
* Let's use `pino` instead.
* We're already using `pino` in other JavaScript files.
* This idea originated from [665](https://github.com/GoogleCloudPlatform/microservices-demo/pull/665/files#r774204647). 

### What purpose/environment will this feature serve? 

* See [pull-request 66](https://github.com/GoogleCloudPlatform/microservices-demo/pull/66) — where the use of `pino` was introduced — for reasoning. 
* Basically, we want our logs to be formatted as JSON — for [Stackdriver](https://cloud.google.com/products/operations). See #47.

NimJay:
## Update (2022 July)
Just adding an update to cool this issue down (i.e., remove it from our team's out-of-SLO list).
No progress updates and this issue it still very low priority.
I wonder if this ticket will be addressed (indirectly) by https://github.com/GoogleCloudPlatform/microservices-demo/issues/422.

aarushisoni:
Hi I want to contribute to this issue please assign me this issue

NimJay:
Hi @aarushisoni,
Thanks again for your interest.
This is definitely a good first issue to work on. :)
I've assigned it to you.

TonyBhargav:
Hi @NimJay since no progress in this issue can I get assigned to this issue. I'll be happy to work on this.
@aarushisoni are you still working on this?

andresmc98:
Hi @NimJay 
is someone working on this issue? i would like give it a shot c:

NimJay:
Hi @andresmc98, :)

I don't think anyone was working on this. Correct my if I'm wrong, @TonyBhargav. :)
I just assigned it to you.

Just a heads-up: please see before starting any work:
[...]

PR #2262, opened 2023-11-13
#674

DracFiendMG:
Use pino instead of console.log


bourgeoisor:
Hey @DracFiendMG! It looks like this branch brought in unrelated commits from years ago.
image

Would you mind recreating your branch from the latest HEAD instead?

DracFiendMG:
Thanks @bourgeoisor, I'll do that right away

PR #2263, opened 2023-11-14
Use pino instead of console.log 

### LINKED ISSUE #674: Use `pino` instead of `console.log`

DracFiendMG:
This fixes #674.


NimJay (changes requested):
@DracFiendMG, thanks so much for starting this!
It looks like `currencyservice` is failing to build because the `logger` object hasn't been initialized inside `server.js`.

```
      > [currencyservice-6f5d5cf7b6-vwlmb server] /usr/src/app/server.js:18
      > [currencyservice-6f5d5cf7b6-vwlmb server]   logger.info("Profiler disabled.")
      > [currencyservice-6f5d5cf7b6-vwlmb server]   ^
      > [currencyservice-6f5d5cf7b6-vwlmb server] ReferenceError: Cannot access 'logger' before initialization
      > [currencyservice-6f5d5cf7b6-vwlmb server]     at Object.<anonymous> (/usr/src/app/server.js:18:3)
```

You may want to test your changes locally by deploying Online Boutique to a [kind](https://kind.sigs.k8s.io/) (or [minikube](https://minikube.sigs.k8s.io/docs/start/)) Kubernetes cluster on your machine.
You can set this up by following [/docs/development-guide.md](https://github.com/GoogleCloudPlatform/microservices-demo/blob/main/docs/development-guide.md)

DracFiendMG:
Sure @NimJay, I'll do that and get back to you

DracFiendMG:
Hi @NimJay, I've deployed and tested this locally and also committed the relevant changes, can you check it now?

And I have one question: When I did **skaffold run** i noticed that almost 15-20 GB was occupied on my disk, do you know where these files are getting stored? I was not able to find them

NimJay:
@DracFiendMG, you asked:
> almost 15-20 GB was occupied on my disk, do you know where these files are getting stored?

That's likely a result of the Docker container images that are stored locally on your machine — when you run `skaffold run`.

You can list Docker images and containers on your machine using:
```
docker image ls
```

You can **delete** images and containers using:
```
docker rmi $(docker images -a -q) -f
docker rm $(docker ps -a -q) -f
```

DracFiendMG (commented):
I've initialized pino as per the requirements and updated the branch as well

src/currencyservice/server.js
```diff
@@ -14,11 +14,22 @@
...
+const logger = pino({
+  name: 'currencyservice-server',
+  messageKey: 'message',
+  formatters: {
+    level (logLevelString, logLevelNum) {
+      return { severity: logLevelString }
+    }
+  }
+});
+
```
NimJay, line 27: **Suggestion:** (feel free to ignore)
Feel free to create a separate `logger.js` file (in each Node.js microservice) where we initialize the `logger = pino(...)` object (just once for each microservice) and re-use that object in other files (by importing the object from the `logger.js` file) but that's a bonus (out-of-scope) for this pull-request.

src/paymentservice/index.js
```diff
@@ -69,4 +69,15 @@ const PROTO_PATH = path.join(__dirname, '/proto/');
...
+const logger = pino({
+  name: 'paymentservice-server',
+  messageKey: 'message',
+  formatters: {
+    level (logLevelString, logLevelNum) {
+      return { severity: logLevelString }
+    }
+  }
+});
+
```
NimJay, line 82: **Issue:** The paymentservice is failing deployment:
```
... deployment/paymentservice: container server terminated with exit code 1 ...
... /usr/src/app/index.js:21
... logger.info("Profiler disabled.")
... ReferenceError: Cannot access 'logger' before initialization
```
You will need to move the chunk that initializes the `logger` object to the top of the file (after `'use strict';`).

NimJay:
Thanks for your continual progress on this change, @DracFiendMG! :)
I left 2 more comments — only one issue needs to be fixed.

src/paymentservice/index.js
DracFiendMG: Sorry about that @NimJay, actually I already initialized logger at the top of the file and tested it out locally but I forgot to commit those changes to my branch.

src/currencyservice/server.js
DracFiendMG: Yeah, in the beginning I was going to do it this way, but creating multiple logger files for different services seemed a bit too much so I only made changes in the required services. 
If you have a suggestion where we only use one `logger.js` and that can be used across all the services then do tell me about it. Because the only change in the logger is the name of the service, if we know how to set the name of the logger's service dynamically then it should be good. 
If that's not possible then maybe you can raise another issue for this and I'll create separate `logger.js` files for all the services.

DracFiendMG: For now, I've created `logger.js` for paymentservice only.
Please check and validate that

DracFiendMG:
@NimJay, can you look into this `header-check` issue, I don't know about this

src/shared/logger.js
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright 2018 Google LLC
```
NimJay, line 2: First off, thank you for creating a separate `logger.js` file!
The `header-check` is failing before the year should be `2023` on new files introduced by this pull-request.
```suggestion
 * Copyright 2023 Google LLC

```

src/currencyservice/server.js
NimJay: Thanks for going the extra mile (but know that you don't _have_ do)! The `logger.js` file of the paymentservice should live inside `/src/paymentservice`.

src/shared/logger.js
DracFiendMG: I've resolved this one

DracFiendMG:
@NimJay, I've performed the suggested changes.
I think it won't fail this time

src/paymentservice/logger.js
```diff
@@ -0,0 +1,27 @@
...
+
+module.exports = pino({
+  name: 'paymentservice-server',
+  messageKey: 'message',
+  formatters: {
+    level (logLevelString, logLevelNum) {
+      return { severity: logLevelString }
+    }
+  }
+});
```
NimJay, line 27: **Thought:** We should use this in [`src/paymentservice/charge.js`](https://github.com/GoogleCloudPlatform/microservices-demo/blob/8bf9151b3fe6014cb3683bf39bcd02d6c418df63/src/paymentservice/charge.js#L19C7-L19C13) as well.

NimJay (approved):
Nice, it works!
@DracFiendMG, thanks so much for your patience with this pull-request.

Approved!
Remaining work:
1. Introduce a `logger.js` file in currencyservice.
2. Update [`src/paymentservice/charge.js`](https://github.com/GoogleCloudPlatform/microservices-demo/blob/8bf9151b3fe6014cb3683bf39bcd02d6c418df63/src/paymentservice/charge.js#L19C7-L19C13) to also use `src/paymentservice/logger.js `.

But let's do those things outside of the pull-request.

DracFiendMG:
@NimJay Thank you, I loved doing this~
If there's any other issues that I may be able to take up then please mention those

PR #2275, opened 2023-11-28
feat: Add support for packaging service

NimJay:
* As part of an upcoming demo, Online Boutique (Cymbal Shops) needs to make a request to a new microservice called the packaging service. Developers will create the packaging service as part of the demo — so its source code likely won't be in this repo.
* This pull-request adds a feature to Online Boutique: developers can set a `PACKAGING_SERVICE_URL` environment variable (in the frontend Pod). Setting this environment variable will make the frontend request and display packaging info inside individual product pages.
* For more context, see [Google-internal link](http://go/duet-ai-demo-for-ga-launch-draft), or ping me for more info.

---

#### **1.** If the `PACKAGING_SERVICE_URL` is not set, no packaging info is shown.
* No change is made to the default deployment of Online Boutique.

Screenshot 2023-11-28 at 3 42 58 PM

#### **2.** If `PACKAGING_SERVICE_URL` is set but the request fails.
* I've tested this scenario inside [pull-request 2277](https://github.com/GoogleCloudPlatform/microservices-demo/pull/2277). See [staging url from that pull-request](http://35.225.40.12/product/1YMWWN1N4O).

Screenshot 2023-11-28 at 3 42 58 PM

#### **3.** If `PACKAGING_SERVICE_URL` is set and the request to the packaging service is successful.
* I've tested this scenario inside [pull-request 2277](https://github.com/GoogleCloudPlatform/microservices-demo/pull/2277). See [staging url from that pull-request](http://35.225.40.12/product/2ZYFJ3GM2N).

Screenshot 2023-11-28 at 3 46 21 PM


src/frontend/packaging_info.go
```diff
@@ -0,0 +1,66 @@
...
+	Width  float32 `json:"width"`
+	Height float32 `json:"height"`
+	Depth  float32 `json:"depth"`
+}
+
+func httpGetPackagingInfo(productId string) (PackagingInfo, error) {
+	// Make the GET request
+	url := os.Getenv("PACKAGING_SERVICE_URL") + productId
+	fmt.Println("Requesting packaging info from URL: ", url)
+	resp, err := http.Get(url)
```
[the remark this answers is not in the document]
NimJay: I've marked this warning as "These alerts are not in production code". This GitHub repo is a sample/demo app. Plus, this URL will — inevitably — be set by the user that deploys the app (it's part of a demo we're supporting).

src/frontend/handlers.go
```diff
@@ -189,21 +189,33 @@ func (fe *frontendServer) productHandler(w http.ResponseWriter, r *http.Request)
...
+		"product":                      product,
+		"recommendations":              recommendations,
+		"cart_size":                    cartSize(cart),
+		"platform_css":                 plat.css,
+		"platform_name":                plat.provider,
+		"is_cymbal_brand":              isCymbalBrand,
+		"deploymentDetails":            deploymentDetailsMap,
+		"frontendMessage":              frontendMessage,
+		"packagingInfo":                packagingInfo,
+		"isPackagingServiceConfigured": isPackagingServiceConfigured(),
```
NimJay, line 218: Sorry, Golang's auto-formatting confused this diff:
All I'm adding:
```
		"packagingInfo":                packagingInfo,
		"isPackagingServiceConfigured": isPackagingServiceConfigured(),
```

bourgeoisor: Is `isPackagingServiceConfigured` necessary if e.g. `packagingInfo` would be null or empty if it's not configured (thus you could do `if packageInfo != null` for example)? Or can `packagingInfo` be null or empty even if packaging is configured / enabled configured?

kubernetes-manifests/frontend.yaml
```diff
@@ -88,6 +88,8 @@ spec:
           #   value: "true"
           # - name: FRONTEND_MESSAGE
           #   value: "Replace this with a message you want to display on all pages."
+          - name: PACKAGING_SERVICE_URL
+            value: "" # This value should look like "http://123.456.789/"
```
bourgeoisor, line 92: Just that comment might be confusing to users. "What is the packaging service url? Do I need this? Where do I find more information about it?"

bourgeoisor (commented):
Were you planning on merging this to latest? Or a separate branch? Or something else entirely?

kubernetes-manifests/frontend.yaml
NimJay: I've updated the comment here, and also commented out the YAML fields.
```
          # As part of a Google Cloud demo, you can run an optional microservice called the "packaging service".
          # - name: PACKAGING_SERVICE_URL
          #   value: "" # This value would look like "http://123.456.789/"
```


Unfort, there is no public documentation for this demo yet (that we can reference here).

src/frontend/handlers.go
NimJay: Good question.

I've updated the code the use _**pointers**_ to the PackagingInfo struct:
```
var packagingInfo *PackagingInfo = nil
```
and
```
func httpGetPackagingInfo(productId string) (*PackagingInfo, error) {
```

The HTML template no longer uses a `isPackagingServiceConfigured` boolean.

NimJay:
@bourgeoisor aksed:

> Were you planning on merging this to latest? Or a separate branch? Or something else entirely?

Yes, I am hoping to merge this into main.
I'll also follow-up with a new release of Online Boutique so these changes are available in our latest images.

src/frontend/packaging_info.go
```diff
@@ -0,0 +1,70 @@
...
+
+func isPackagingServiceConfigured() bool {
+	return os.Getenv("PACKAGING_SERVICE_URL") != ""
+}
+
+func httpGetPackagingInfo(productId string) (*PackagingInfo, error) {
+	// Make the GET request
+	url := os.Getenv("PACKAGING_SERVICE_URL") + productId
+	fmt.Println("Requesting packaging info from URL: ", url)
+	resp, err := http.Get(url)
```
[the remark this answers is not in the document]
minherz: @NimJay I agree with the scanner :slightly_smiling_face: _ANY_ URL can be placed here. While it is up to the cluster admin what the value would be but can we improve it a little by minimizing the attach surface? For example, can we limit URLs only to the cluster IPs or do another kind of validation?

kubernetes-manifests/frontend.yaml
```diff
@@ -88,6 +88,9 @@ spec:
           #   value: "true"
           # - name: FRONTEND_MESSAGE
           #   value: "Replace this with a message you want to display on all pages."
+          # As part of a Google Cloud demo, you can run an optional microservice called the "packaging service".
```
minherz, line 91: nit: Placing it into main branch takes it beyond Google Cloud demo. Do we want to make it optional or change the comment here?

kubernetes-manifests/frontend.yaml
```diff
@@ -88,6 +88,9 @@ spec:
...
+          # - name: PACKAGING_SERVICE_URL
+          #   value: "" # This value would look like "http://123.456.789/"
```
minherz, line 93: nit: The comment contain invalid IPv4 address. Is it intentional?

src/frontend/packaging_info.go
```diff
@@ -0,0 +1,70 @@
...
+
+type PackagingInfo struct {
+	Weight float32 `json:"weight"`
+	Width  float32 `json:"width"`
+	Height float32 `json:"height"`
+	Depth  float32 `json:"depth"`
+}
+
+func isPackagingServiceConfigured() bool {
+	return os.Getenv("PACKAGING_SERVICE_URL") != ""
```
minherz, line 38: nit: what is a reasoning to read env variable each time this method is called? Is there a way to update the environment variable without restarting the service?

minherz, line 38: nit: should we validate that it is a valid URL and not just some string with non-zero length?

src/frontend/packaging_info.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func httpGetPackagingInfo(productId string) (*PackagingInfo, error) {
+	// Make the GET request
+	url := os.Getenv("PACKAGING_SERVICE_URL") + productId
```
minherz, line 43: nit: it is a second place which reads the env variable on each call. if there is no way to update the environment without restarting the service, please consider caching this value.

src/frontend/handlers.go
```diff
@@ -189,6 +189,16 @@ func (fe *frontendServer) productHandler(w http.ResponseWriter, r *http.Request)
...
 
+	// Fetch packaging info (weight/dimensions) of the product
+	// The packaging service is an optional microservice you can run as part of a Google Cloud demo.
+	var packagingInfo *PackagingInfo = nil
+	if isPackagingServiceConfigured() {
+		packagingInfo, err = httpGetPackagingInfo(id)
+		if err != nil {
+			fmt.Println("Failed to obtain product's packaging info:", err)
+		}
+	}
```
minherz, line 200: nit: if `isPackagingServiceConfigured` is never used except in conjunction with ` httpGetPackagingInfo`, what is a reason having it as a stand alone function? consider doing this check inside `httpGetPackagingInfo` and processing two errors: http vs not defined

src/frontend/packaging_info.go
```diff
@@ -0,0 +1,70 @@
...
+	fmt.Println("Requesting packaging info from URL: ", url)
```
minherz, line 44: nit: Consider validating the resulted URL

src/frontend/packaging_info.go
minherz, line 45: nit: if frontend has any retry mechanism when calling other services, it worth adding it to this call as well

kubernetes-manifests/frontend.yaml
NimJay: I've updated `123.456.789` to `123.123.123`.

src/frontend/packaging_info.go
NimJay: Good thought. Since the frontend doesn't already have such a "retry mechanism", I'll let this be.

src/frontend/packaging_info.go
NimJay: I agree that some validation would be ideal.
But I will skip this for now because:
1. For the demo, we might end up using K8s service addresses (`namespace.packagingservice`) instead of a public IP addresses (`http://123.123.123`).
2. We don't do such validation for other environment variables.

NimJay: I agree that some validation would be ideal.
But I will skip this for now because:
1. For the demo, we might end up using K8s service addresses (`namespace.packagingservice`) instead of a public IP addresses (`http://123.123.123`).
2. We don't do such validation for other environment variables.

src/frontend/packaging_info.go
NimJay: I agree that some validation would be ideal.
But I will skip this for now because:
1. For the demo, we might end up using K8s service addresses (`namespace.packagingservice`) instead of a public IP addresses (`http://123.123.123`).
2. We don't do such validation for other environment variables.

src/frontend/packaging_info.go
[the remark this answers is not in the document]
NimJay: The demo will either use a:
1. public IP address for the packaging service
2. K8s `Service` address of the packaging service (`default.packagingservice`) — if the new packaging service ends up being in the same cluster (e.g., we might use Cloud Run instead of GKE for the packaging service)

NimJay:
Merged without approval since this repo currently has no team officially owning it — plus: 
* all comments from reviewer feedback have been address,
* checks are passing, and
* this new feature has been tested via https://github.com/GoogleCloudPlatform/microservices-demo/pull/2277.

PR #2323, opened 2024-01-05
feat: add 'app' label to Deployment and Service objects

muncus:
### Background 

When searching by the 'app' label, only pods would appear. To find the Service
or Deployment objects associated with the app, one would need to query by name.
In some cases (that is, 'frontend-external'), name is not sufficient.

Some tools, like the backstage.io kubernetes plugin, do not show the Deployment
and Service objects because they are missing these labels. This leads to
confusion about the service/deployment that owns the pods.

### Change Summary

Adds the 'app' label to all Service and Deployment objects

### DIFF
--- a/kubernetes-manifests/adservice.yaml
+++ b/kubernetes-manifests/adservice.yaml
@@ -16,6 +16,8 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: adservice
+  labels:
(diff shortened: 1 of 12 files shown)


NimJay (approved):
Looks great! Thanks for catching and fixing this, @muncus!
Approved. Merging.

PR #2389, opened 2024-02-25
feat: strengthen the load generator with fuzz input

Lincyaw:
### Background 

The original load generator only sends the same request template, which lacks diversity.
Adding a fuzzy-style request generator will make the request more general.


src/loadgenerator/requirements.txt
```diff
@@ -70,6 +70,7 @@ zope-event==5.0
     # via gevent
 zope-interface==6.1
     # via gevent
-
+faker==23.2.1
+    # via faker
```
NimJay, line 74: **Issue:** This `faker` import should be specified inside [`src/loadgenerator/requirements.in`](https://github.com/GoogleCloudPlatform/microservices-demo/blob/a1240bfa16bf308b58280e506880a1090a262013/src/loadgenerator/requirements.in).

src/loadgenerator/locustfile.py
```diff
@@ -47,22 +50,30 @@ def addToCart(l):
     l.client.get("/product/" + product)
     l.client.post("/cart", {
         'product_id': product,
-        'quantity': random.choice([1,2,3,4,5,10])})
+        'quantity': random.randint(1,100)})
+    
+def empty_cart(l):
```
NimJay, line 55: **Thought**:
Other method names use camelCase, but `empty_cart` is good.
We should use snake_case for Python function names, based on this [StackOverflow post](https://stackoverflow.com/a/8908798).
No action needed (since this is out-of-scope for your change and you've gotten it right).

NimJay (changes requested):
Thank you so much for this improvment, @Lincyaw! It look great!
There's just one issue regarding the `requirements.in` file — see my comment(s).

src/loadgenerator/locustfile.py
NimJay: I've opened https://github.com/GoogleCloudPlatform/microservices-demo/issues/2397. No action needed from this pull-request.

Lincyaw:
Thanks for your review! I will open another pull request

PR #2398, opened 2024-02-27
feat: strengthen the load generator with fuzz input

Lincyaw:
Background
The original load generator only sends the same request template, which lacks diversity.
Adding a fuzzy-style request generator will make the request more general.


NimJay (approved):
Thank you again, @Lincyaw, for this improvement.
And thanks for addressing [my feedback from your previous pull-request](https://github.com/GoogleCloudPlatform/microservices-demo/pull/2389#pullrequestreview-1901112644).

I'm just waiting on the tests to run now.
Your changes look good. Approved.

PR #2410, opened 2024-03-08
fix: remove potentially unsafe external link

iennae:
### Background 
this updates the attributes associated with the anchor tag and addresses CWE-200 and CWE-1022 

### Fixes 
HTML links that open in a new tab or window allow the target page to access the DOM of the origin page using window.opener unless link type noopener or noreferrer is specified. This is a potential security risk.

### Change Summary
rel="noopener": This attribute enhances security. When a link opens in a new tab (often with target="_blank"), noopener prevents the newly opened webpage from accessing the window object of the original page. This mitigates a security vulnerability called "tabnabbing" where a malicious website could potentially steal information from the original webpage.

rel="noreferrer": Similar to noopener, this prevents the destination website from knowing that the link came from your webpage. Browsers typically send referrer information along with link clicks, which can be used by the destination site to track where their traffic originates from. With noreferrer, this information is withheld.

### DIFF
--- a/src/frontend/templates/ad.html
+++ b/src/frontend/templates/ad.html
@@ -18,7 +18,7 @@
 <div class="container py-3 px-lg-5 py-lg-5">
     <div role="alert">
         <strong>Ad</strong>
-        <a href="{{.RedirectUrl}}" rel="nofollow" target="_blank">
+        <a href="{{.RedirectUrl}}" rel="nofollow noopener noreferrer" target="_blank">
             {{.Text}}
         </a>
     </div>


NimJay (approved):
Thanks so much for this fix, @iennae! Looks good.

(I only just learned about `noopener` and `noreferrer` through your pull-request. This is good info.)

Today, the value at `href="{{.RedirectUrl}}"` is always a link to Online Boutique itself ([see search](https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fmicroservices-demo%20RedirectUrl&type=code)), so this vulnerability has not been exploited. 😌
But that's not guaranteed (but _should_ be guaranteed) at the level of this `src/frontend/templates/ad.html` file so this is a good add.

Aside: The GitHub check that deploys the staging URL is currently broken, but this should be safe to merge (without testing deployment). Merging.

PR #2424, opened 2024-03-15
Only run PR CI on most recent commit

NimJay:
### Background 
* Today, the `code-tests` and `deployment-test` GitHub Actions jobs get queued up for _all_ commits of a pull-request.
* But we only care about running these jobs for the most recent commit of a pull-request.

### Change Summary
* In this pull-request, I'm making `.github/workflows/ci-pr.yaml` cancel a queued-up workflows if it's not the workflow isn't running on the most recent commit.
* See [this StackOverflow answer for an explanation](https://stackoverflow.com/a/72408109).

### Testing Procedure
* You'll see that it worked for the commits in this pull-request:

image

### DIFF
--- a/.github/workflows/ci-pr.yaml
+++ b/.github/workflows/ci-pr.yaml
@@ -59,6 +59,7 @@ jobs:
       timeout-minutes: 10
       run: |
         dotnet test src/cartservice/
+
   deployment-tests:
     runs-on: [self-hosted, is-enabled]
     needs: code-tests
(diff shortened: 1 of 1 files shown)


bourgeoisor (approved):
LGTM!

NimJay:
Thanks for the super speedy review/approval, Olivier. ⚡ 

Looks like this change is doing it's thing:
image

PR #2425, opened 2024-03-15
AlloyDB support for the products catalog

bourgeoisor:


src/productcatalogservice/server.go
```diff
@@ -113,6 +116,88 @@ func main() {
 		}
 	}()
 
+	// TODO
+	projectID := "obourgeois-sandbox-13"
```
moficodes, line 120: We are hardcoding credentials/projects here. probably should not push this?

bourgeoisor: There's no useful creds in that branch, the process needs an SA key to talk to anything, which is not checked in.
But yeah hard-coding is not permanent, just wanted to get "something" working (and not quite there yet).

PR #2467, opened 2024-04-08
Fix credit card validation error

NimJay:
### Background 
* Current, checking out produces the following error:
Screenshot 2024-04-08 at 2 04 03 PM

### Change Summary
* Remove dashes ("-"s) in credit card during submission.

### Testing Procedure
* We just need to test that checking out (via the staging URL) works.


bourgeoisor (approved):
Staging deployment works as intended. Looks good to me!

PR #2472, opened 2024-04-10
Cleanup README

bourgeoisor:
This PR cleans up the README and puts it more in-line in terms of consistency with [Bank of Anthos](https://github.com/GoogleCloudPlatform/bank-of-anthos):
- Added list of Google Cloud products (with links to docs)
- Consolidated list of non-quickstart deployment options under one header
- Removed redundant list of features
- Bumped the architecture at the top

Staged: https://github.com/GoogleCloudPlatform/microservices-demo/blob/issue-cleanup-readme/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,42 @@
-<p align="center">
+<!-- <p align="center">
 <img src="/src/frontend/static/icons/Hipster_HeroLogoMaroon.svg" width="300" alt="Online Boutique" />
-</p>
-
+</p> -->
 ![Continuous Integration](https://github.com/GoogleCloudPlatform/microservices-demo/workflows/Continuous%20Integration%20-%20Main/Release/badge.svg)
 
 **Online Boutique** is a cloud-first microservices demo application.  The application is a
-web-based e-commerce app where users can browse items,
-add them to the cart, and purchase them.
+web-based e-commerce app where users can browse items, add them to the cart, and purchase them.
 
-Google uses this application to demonstrate the use of technologies like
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,15 +65,8 @@ If you’re using this demo, please **★Star** this repository to show your int
    ```
 
    Substitute `<PROJECT_ID>` with the ID of your Google Cloud project.
-  
-4. Confirm the services have been enabled for your project.
```
bourgeoisor, line 50: Noting that I'm removing this redundant step, since the step above is not async and would tell you immediately if there was an error.

README.md
```diff
@@ -1,21 +1,40 @@
-<p align="center">
+<!-- <p align="center">
```
bourgeoisor, line 1: Noting that I'm commenting this image since it added quite a bit of extra scroll for the user to get to the README / docs / content. We can always re-add it at any time if we feel it should be there.

bourgeoisor:
cc: @NimJay merging this primarily due to the influx of traffic from the Next 24 dev keynote for those looking to try out the AI demo (which is nowhere to be found in the README at the moment). We can chat later on about any feedback / changes you'd like to see!

README.md
NimJay: I like this (removal of logo) — especially since:
1. we don't really use this logo anywhere else (i.e., it's not essential)
2. getting straight to the arch diagram is more important

README.md
NimJay: I like — shortening time to completion.

PR #2473, opened 2024-04-10
Refactor common default templating payload in frontend

### LINKED ISSUE #2450: Update copyright to be Time.now.year (or equivalent)
### Describe the bug 
The copyright says 2020 and there's a request to have the current year instead.

bourgeoisor:
This PR extracts all of the default payload data sent to frontend templates into a single injection function, to avoid having to repeat the same data for every handler route.

Also fixes #2450 as a side-effect

## How to test?

View the staged link (http://34.30.20.128), and notice that the pages are working as intended / not broken, and that the copyright footer now says `2020-{CURRENT YEAR}` (in other words, `2020-2024`)

### DIFF
--- a/src/frontend/handlers.go
+++ b/src/frontend/handlers.go
@@ -107,7 +107,7 @@ func (fe *frontendServer) homeHandler(w http.ResponseWriter, r *http.Request) {
 	plat = platformDetails{}
 	plat.setPlatformDetails(strings.ToLower(env))
 
-	if err := templates.ExecuteTemplate(w, "home", injectDefaultData(r, map[string]interface{}{
+	if err := templates.ExecuteTemplate(w, "home", injectCommonTemplateData(r, map[string]interface{}{
 		"show_currency": true,
 		"currencies":    currencies,
 		"products":      ps,
@@ -195,7 +195,7 @@ func (fe *frontendServer) productHandler(w http.ResponseWriter, r *http.Request)
 		}
 	}
 
-	if err := templates.ExecuteTemplate(w, "product", injectDefaultData(r, map[string]interface{}{
+	if err := templates.ExecuteTemplate(w, "product", injectCommonTemplateData(r, map[string]interface{}{
 		"ad":              fe.chooseAd(r.Context(), p.Categories, log),
 		"show_currency":   true,
 		"currencies":      currencies,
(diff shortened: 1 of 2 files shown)


src/frontend/templates/footer.html
```diff
@@ -20,7 +20,7 @@
     <div class="footer-top">
         <div class="container footer-social">
             <p class="footer-text">This website is hosted for demo purposes only. It is not an actual shop. This is not a Google product.</p>
-            <p class="footer-text">© 2020 Google Inc (<a href="https://github.com/GoogleCloudPlatform/microservices-demo">Source Code</a>)</p>
+            <p class="footer-text">© 2020-{{ .currentYear }} Google LLC (<a href="https://github.com/GoogleCloudPlatform/microservices-demo">Source Code</a>)</p>
```
NimJay, line 23: TBH, I would have removed this "copyright" footer text altogether because it doesn't seem to be a requirement + our CSS/HTML/frontend source code itself contains copyrights + we should remove as much complexity as possible from Online Boutique.

But this works too.

src/frontend/handlers.go
```diff
@@ -584,19 +539,34 @@ func renderHTTPError(log logrus.FieldLogger, r *http.Request, w http.ResponseWri
...
+	if templateErr := templates.ExecuteTemplate(w, "error", injectDefaultData(r, map[string]interface{}{
+		"error":       errMsg,
+		"status_code": code,
+		"status":      http.StatusText(code),
+	})); templateErr != nil {
+		log.Println(templateErr)
+	}
+}
+
+func injectDefaultData(r *http.Request, payload map[string]interface{}) map[string]interface{} {
```
NimJay, line 551: Love this! 
**Suggestion:** But let's rename this function so that it's more clear.
Some options:
- `injectDefaultTemplateData`
- `injectCommonTemplateData`

(`data` is definitely the correct word here — based on [the documentation](https://pkg.go.dev/html/template#Template.ExecuteTemplate).)

NimJay (approved):
Thanks, @bourgeoisor! This looks great.
Just left one comment that needs resolving (renaming to something like `injectCommonTemplateData`).

My manual test of the staging URL worked. 👍  Approved.

(Feel free to merge yourself after resolving my suggestion + doing a final manual test of the staging URL yourself.)

src/frontend/templates/footer.html
bourgeoisor: Agreed but I didn't want to break the flow too much. This PR does have a net removal of 30 lines of code though, so that's a win :)

src/frontend/handlers.go
bourgeoisor: Oh yeah that makes more sense. Something about "naming things" being a hard problem :)

PR #2578, opened 2024-06-06
fix(kustomize/without-loadgenerator): update wrong patch struct

aslafy-z:

### DIFF
--- a/kustomize/components/without-loadgenerator/kustomization.yaml
+++ b/kustomize/components/without-loadgenerator/kustomization.yaml
@@ -15,4 +15,4 @@
 apiVersion: kustomize.config.k8s.io/v1alpha1
 kind: Component
 patches:
-- delete-loadgenerator.patch.yaml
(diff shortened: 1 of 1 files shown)


NimJay (approved):
@aslafy-z, thanks so much for this fix and the super clear pull-request description!

Strange, I didn't encounter that error (`Error: invalid Kustomization: json: cannot unmarshal string into Go struct field Kustomization.patches of type types.Patch`) in either my local or our GitHub checks.

The kustomize and kubectl versions I'm using locally:
```
> kustomize version
{Version:kustomize/v4.5.7 GitCommit:56d82a8378dfc8dc3b3b1085e5a6e67b82966bd7 BuildDate:2022-08-02T16:28:01Z GoOs:darwin GoArch:amd64}

> kubectl version
Client Version: v1.29.5
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
...
```

NimJay (approved):

PR #2598, opened 2024-06-18
`adservice` - `ca-certificates` is already in the final base image

mathieu-benoit:
No need to add `ca-certificates` nor adding a container layer for this `RUN` statement since this package is already in the base container image (`eclipse-temurin:jre-alpine`).

Some space on disk is also saved, from `258 MB` to `230 MB` (-`28 MB`).


mathieu-benoit:

bourgeoisor (approved):

PR #2614, opened 2024-07-02
`bootstrap@5.3.3`

mathieu-benoit:
`bootstrap@5.3.3` for `frontend`

https://getbootstrap.com/docs/versions/

### DIFF
--- a/src/frontend/templates/footer.html
+++ b/src/frontend/templates/footer.html
@@ -47,9 +47,7 @@
         </div>
     </div>
 </footer>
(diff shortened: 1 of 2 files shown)


mathieu-benoit:
Tested in there http://35.238.70.73/, ready for your review, thanks!

mathieu-benoit:
Actually it seems that there is an issue in the layout of `Quantity` in the /cart page.

Before:
image


After:
image


We could see that `Quantity : 1                              $24.99` is floating nowhere in the second image, while in the first image the associated `Quantity: 1                              $18.99` is well embedded in the `row cart-summary-item-row`.

Not mergeable yet.

I'm not a FE dev, anyone would like to look at this?

PR #2636, opened 2024-07-19
feat(helm): adding property to define annotations on external service that would enable integration with external-dns

justinas-b:
adding property to define annotations on external service that would enable integration with external-dns


justinas-b:
@yoshi-approver could you please take a look?

PR #2758, opened 2024-10-02
Add optin for securitycontext in helm

sandert-k8s:
### Background 
This issue resolves that the app won't be able to be deployed on OpenShift. OpenShift has it's own securityContext mechanism (different UserId's per Namespace/Project). When using other userId's than provided by OpenShift, the app won't start.
The default value of the variable is true, so default it will use the securityContext from the chart.

### DIFF
--- a/helm-chart/values.yaml
+++ b/helm-chart/values.yaml
@@ -210,6 +210,9 @@ cartDatabase:
     endpointPort: ""
     certificate: ""
 
+securityContext:
+  enable: false
+
 # @TODO: This service is not currently available in Helm.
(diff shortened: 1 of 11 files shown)


helm-chart/values.yaml
```diff
@@ -210,6 +210,9 @@ cartDatabase:
     endpointPort: ""
     certificate: ""
 
+securityContext:
```
bourgeoisor, line 213: Could we move this before the individual services values? So around line 58?

sandert-k8s: Of course!

bourgeoisor (approved):
LGTM

PR #2759, opened 2024-10-03
Add Route to Frontend service

sandert-k8s:
### Background 
Adds a OpenShift Route to the Frontend service. It is disabled by default, and can be enabled by setting the `frontend.route.create` toggle to `true`

If this doesn't suit the project (because it's OpenShift only), I can understand. This PR can be closed then.

### DIFF
--- a/helm-chart/templates/frontend.yaml
+++ b/helm-chart/templates/frontend.yaml
@@ -283,3 +283,24 @@ spec:
           number: 80
 {{- end }}
 {{- end }}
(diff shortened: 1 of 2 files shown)


bourgeoisor:
Hi @sandert-k8s! Yeah unfortunately this one will be closed since it doesn't suit the primary purposes of this repo: https://github.com/GoogleCloudPlatform/microservices-demo/blob/main/docs/purpose.md

Let me know if you have any questions about this.

PR #2806, opened 2024-11-18
Configurable Locust rate

Arri98:
### Background 
The loadgenerator deployment currently uses the default Locust user spawn rate of 1 user per second. This makes it difficult to showcase GKE and Kubernetes autoscaling as the slow spawn rate does not generate sufficient load.

### Change Summary

This PR exposes Locust's rate parameter to:
  - kubernetes-manifests.yaml file.
  - Helm's frontend.yaml template. The value is not added to the values.yaml as the user parameter is also not exposed in this file.

Loadgenerator image entrypoint is modified to parse the rate parameter. Rate is set to the default value for backwards compatibility.

### Testing Procedure
Modify the rate parameter and deploy the updated manifest. The rate parameter will be displayed in the loadgnerator pod logs.


release/kubernetes-manifests.yaml
```diff
@@ -741,6 +741,8 @@ spec:
           value: "frontend:80"
         - name: USERS
           value: "10"
+        - name: RATE
```
bourgeoisor, line 744: Would need to be added to the kustomize base too:
https://github.com/GoogleCloudPlatform/microservices-demo/blob/main/kustomize/base/loadgenerator.yaml#L84

Arri98: Done

bourgeoisor (approved):
LGTM. Note that this will not be available until the next release, unless you manually deploy from the Skaffold (i.e. `skaffold run`)

PR #2813, opened 2024-11-26
Shrink emailservice base image, to python:3.13.0-alpine

NimJay:
### Background 

* `python...alpine` is one of the smallest `python...` images. See list of other `python...` images: https://hub.docker.com/_/python.
* Smaller the base image, the better —> fewer vulnerabilities, and faster build/push/pull time.
* According to [this article](https://medium.com/@arif.rahman.rhm/choosing-the-right-python-docker-image-slim-buster-vs-alpine-vs-slim-bullseye-5586bac8b4c9):
> Alpine takes the security crown with its minimal attack surface and frequent updates.
* See related Google-internal bug: b/378279353

### Change Summary
* This change shrinks the base image used by the `emailservice` (from `python:3.12.6-slim`).

### Testing Procedure
* The testing performed by our integration tests (in this PR's GitHub Actions) should suffice.

### DIFF
--- a/src/emailservice/Dockerfile
+++ b/src/emailservice/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM python:3.12.6-slim@sha256:ad48727987b259854d52241fac3bc633574364867b8e20aec305e6e7f4028b26 AS base
+FROM python:3.13.0-alpine@sha256:ad48727987b259854d52241fac3bc633574364867b8e20aec305e6e7f4028b26 AS base
 
 FROM base AS builder
 


src/emailservice/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM python:3.12.6-slim@sha256:ad48727987b259854d52241fac3bc633574364867b8e20aec305e6e7f4028b26 AS base
+FROM python:3.13.0-alpine@sha256:ad48727987b259854d52241fac3bc633574364867b8e20aec305e6e7f4028b26 AS base
```
NimJay, line 15: Looks like I forgot to update the SHA256 hash, so it didn't actually update the image:
```suggestion
FROM python:3.13.0-alpine@sha256:96530606ce7cd1dbcf0baa1e8f1947ec76e16ac260dad8bcd68c27c73cada414 AS base
```

src/emailservice/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM python:3.12.6-slim@sha256:ad48727987b259854d52241fac3bc633574364867b8e20aec305e6e7f4028b26 AS base
+FROM python:3.13.0-alpine@sha256:96530606ce7cd1dbcf0baa1e8f1947ec76e16ac260dad8bcd68c27c73cada414 AS base
```
NimJay, line 15:
```suggestion
FROM python:3.13.0-alpine@sha256:6fe2f8718e257d8be95f79e3ad3659183f01b0b28fe1d50f5f6e3c54ebbd8f12 AS base
```

src/emailservice/Dockerfile
```diff
@@ -12,14 +12,13 @@
...
 
 FROM base AS builder
 
-RUN apt-get -qq update \
-    && apt-get install -y --no-install-recommends \
-        wget g++ \
-    && rm -rf /var/lib/apt/lists/*
+RUN apk update \
+    && apk add --no-cache wget g++ \
+    && rm -rf /var/cache/apk/*
```
NimJay, line 21: * `alpine` uses `apk`, not `apt-get`.
* The `rm -rf /var/cache/apk/*` line removes package lists and information that apt-get downloads during updates. (It might be redundant for our Dockerfile because we're using a multi-stage build, but I'll leave it for now.)

PR #3370, opened 2026-05-19
fix(compose): pull cartservice upstream instead of building

malston:
## Summary

- `docker compose build` for cartservice fails on Apple Silicon with `MSB4184` (BuildKit restore) or protoc SIGSEGV (classic builder publish) under the .NET 10 SDK. Same wall the `build-images.sh` path hit on `fix/build-script`.
- Drop the cartservice `build:` directive from `docker-compose.yml` and remove the cartservice override from `docker-compose.ecr.yml`; both files still reference `cartservice:5096a85`.
- Document a one-time upstream pull + tag step in `WORKSHOP.md` (digest-pinned to `us-central1-docker.pkg.dev/google-samples/microservices-demo/cartservice:v0.10.5@sha256:85c4fee0…`). The pulled image is amd64 and runs under Docker Desktop emulation alongside the locally-built arm64 services.

Workshop labs don't modify cartservice internals, so the version skew is intentional.

## Test plan

- [ ] On Apple Silicon: run the documented `docker pull` + `docker tag` step, then `docker compose -f docker-compose.yml -f docker-compose.ecr.yml build --no-cache` completes without touching cartservice.
- [ ] `docker compose up -d` brings up all 11 services; `cartservice` resolves to the locally-tagged image and is healthy.
- [ ] On amd64 Linux: same flow works (pull is still amd64-native, no emulation needed).
- [ ] Frontend at http://localhost:8080 loads and cart interactions succeed end-to-end.


malston:
Sorry, wrong target -- this is a workshop-internal fix for forgd-ai/microservices-demo, not for upstream. Closing.

PR #3421, opened 2026-07-07
Migrate to GitHub Action runners

olivi-eh:
This PR migrates away from using self-hosted runners.

Instead, we're now using secured GitHub-hosted Actions leveraging Workload Identity to impersonate a Google Cloud Service Account that has very restricted permissions (the ability to deploy k8s resources on the PRs cluster, and the ability to write new Artifact Registry images).

**How to test?**

Check out the attached `Continuous Integration - Pull Request / deployment-tests` run that it succeeded, and the commented staged IP address.


NimJay (approved):
Awesome! Thanks for doing this.
(I assume we no longer need that `addlicense` dependency.)

Ready for merge.

